PR #136, opened 2015-03-13
Indicate re_match uses ERE as well. Please verify.

karloluiten:
From testing a regular expression it seemed re_match uses ERE (as does re_extract). Please verify if this is indeed true and then add it to the documentation so others can easily find it.


rgerhards:
Indeed it does -- thanks!

PR #143, opened 2015-04-15
Documented new action param (action.copyMsg) + some minor restructuring

janmejay:
- Added doc-fragment for action.copyMsg
- Dedup: Got basic-structure page to point to control-structures page from RainerScript section, which talks about control-flow structures in more detail.
- Dedup: Got basic-structure page to point to set/unset/reset documentation in RainerScript section, which talks about mutating statements in much more detail.


janmejay:
Goes with https://github.com/rsyslog/rsyslog/pull/303

PR #185, opened 2015-11-20
Update certtool request per latest version suggestions

kkirsche:
When running this today you get the following notice:

```shell
example@hostname:/etc/rsyslog.d# certtool --generate-privkey --outfile key.pem --bits 4096
** Note: Please use the --sec-param instead of --bits
Generating a 4096 bit RSA private key...
```

For this reason, I am suggesting the move to `--sec-param` to match the suggestion of the tool.


radu-gheorghe:
Merged. Thanks!

PR #188, opened 2015-12-03
Big restructuring

ribeiropaulor:
Preview of first steps of big restructuring: 

http://rsyslog-doc.readthedocs.org/en/big_restructuring/proposals/big_restructuring/index.html


rgerhards:
sorry, I am currently totally swamped and will recover only after the holiday season. I had a quick glimpse and from that I find it promising. As a general comment we need to make sure we don't loose any doc that was written after the big restructuring began -- I remember we had this problem with a similar effort and it bite us quite a bit in the months after (and no, I am not saying that I saw this problem, just a general comment).

If we are lucky @radu-gheorghe has some time to review.

radu-gheorghe:
It's in the backlog, @rgerhards :)

Whenever this happens, thanks a lot @ribeiropaulor for the effort!

radu-gheorghe:
First of all thanks again @ribeiropaulor for the effort!

I finally got to look over the current content and here are my comments:

Sphinx build shows pretty much the same thing. Also, this is a new branch (big_restructuring) so I'm not sure what you have in mind in terms of integrating this into the official docs. I would suggest we simply replace the current docs with the new ones on the master branch, once we have all the content in. Then we shouldn't be afraid of losing any docs, we have other branches (that are still available over rsyslog.com, like v8-stable). Then I would only be concerned about broken links. Tough luck for those who link against the master branch (oops, I have to update some blog posts :p) but I don't think it OK to have broken links for the v8-stable branch, so I guess we'll need a new branch. Like v8-newdocs, and when v9 comes around we could have only the new docs. Thoughts?

I would try to slim down the version-dependent docs as much as possible, to make them easier to maintain. I think that's one of the main pain-points of the current docs: they're incomplete and not well maintained, so people can't trust them. So I would suggest to:
- pull out the cookbook section, and put stuff from there either in Wiki (unlikely, as everything there is oooold) or in rsyslog.com under Help -> Guides for rsyslog.
- pull out the book part. I'm not 100% sure about this, but I'd say it should be general enough that it would be low-maintenance and could live as a separate section within the Help menu, so we can keep only the configuration part dynamic. This would also mean that, for example, we would give a couple of Rainerscript examples in the book part and then link to the configuration reference for an (updated) list of all possibilities. This implies expanding the configuration reference to include things like filter conditions, property replacer and templates
- pull out the Contributing part. This is even less likely to change over time and should stand out as a separate thing IMO

Again, any comments are welcome for the stuff I wrote above. In general, I think the stucture is spot on in terms of what we should cover and how to break it down (which is what this PR is about :p) so I'm +1 for merging it to this big_restructuring branch, and then we could open separate issues for any changes we might want to make. Starting with higher-level changes like the comments I made above and going down to the content itself. And we could discuss each point in its own issue instead of postponing this merge over N general issues.

@rgerhards and @ribeiropaulor, what do you think?

mostolog:
Any news about this?

rgerhards:
@ribeiropaulor @radu-gheorghe  mea culpa, mea culpa... I have totally screwed up! Finally, after 2 years (!) I come back to this very valuable PR... But at least I frankly admit my failure.

My plan now is to 
1. merge as is
1. address the comments from @radu-gheorghe 
1. add a bit of content
1. hope for some feedback/help from @deoren 
1. see that we can use the new top structure relatively soon
1. extend on that

The steps may occur a bit out of order, especially as I would like to move it to the actual doc root ASAP. If we are extremely lucky @ribeiropaulor would still be interested to help a bit more, but I doubt he is after two years.

@ribeiropaulor this was very well thought out, please accept my sincere apologies for not giving it the proper attention for so long!

rgerhards:
mmhhh... this should have been merged via #410, but as it has at least not the state, I did an explicit merge.

PR #191, opened 2015-12-12
Update omelasticsearch.rst [add journal "dependency"]

jnovack:
Document caveat as found in https://github.com/rsyslog/rsyslog/issues/698.

### DIFF
--- a/source/configuration/modules/omelasticsearch.rst
+++ b/source/configuration/modules/omelasticsearch.rst
@@ -200,6 +200,18 @@ The following sample does the following:
            queue.dequeuebatchsize="300"
            action.resumeretrycount="-1")
 
+**Caveats:**
+
+In order to read and forward any log messages from programs which write directly
+to the system journal (including but not limited to systemd services), the
+following module must be loaded before `omelasticsearch`.
+
+::
+
+    module(load="imjournal")
+
+
+
 
 This documentation is part of the `rsyslog <http://www.rsyslog.com/>`_
 project.


rgerhards:
While I like the idea, that's a general thing one needs to concern. So if we add it here, we need to add it to all module descriptions. Maybe we can find a better place for it?

jnovack:
I appreciate your concern, but I'm not sure how to solve it.  If you follow the caveat, I'm suggesting that `imjournal` is a requirement, not so much a caveat.

Running v8.14.0-1, I do not need `imjournal`, but after installing v8.14.0-6, I need `imjournal` when using `omelasticsearch`.  Can any else prove that this is not the case? (`imjournal` does not need to be loaded before `omelasticsearch` running v8.14.0-6+)?

portante:
@jnovack, so you are saying data from the journal is being send to rsyslog for v8.14.0-1 when you have no `imjournal` loaded in that configuration?  Can you share that configuration?  What is your `/etc/systemd/journal.conf` file have?

And are you using the same configuration files for v8.14.0-6?

Is `imjournal` loaded anywhere in the configuration file at all?

jnovack:
No, I'm saying that my configs are the exact same between v8.14.0-1 and v8.14.0-6.  https://gist.github.com/jnovack/8c3af749305f05b8a977

HOWEVER, within `rsyslog.conf`, I HAVE to load `imjournal` to work.  On v8.14.0-1 I do not. This leads me to believe (potentially falsely) that omelasticsearch is REQUIRING that `imjournal` to work, otherwise nothing gets done.

portante:
@jnovack, okay, you have lots of inputs under the v8.14.0-1 case: imuxsock, imklog, imudp, imtcp.  Can add an actiion:

```
*.* /var/log/rsyslog.debug.log;RSYSLOG_DebugFormat
```

And see where the source of the log messages are coming from?  I am curious to know how systemd is forwarding messages to rsyslog if you don't have `imjournal` loaded.  The reason I ask, is because I don't be `journald` is configured by default to forward to rsyslog vi the socket, and on some of the Fedora and RHEL boxes I have looked at, systemd is configured to listen on the `/dev/log` unix socket.  So perhaps finding out how the working case is actually working might give us a clue as to what is not working with v8.14.0-6.

And I assume the v8.14.0-6 is pulled from the same place you got v8.14.0-1?  If so, I'd like to find out how the two were actually built by looking at their original source RPMs.

jnovack:
I am pulling CentOS 7 packages from //rpms.adiscon.com/v8-stable/

I will need time to re-set up a box with 8.14.0-1, I've migrated production to 8.14.0-6 now that I was shown the light of `imjournal`.

rgerhards:
closing this PR due to concerns voiced in my first comment.

jnovack:
This was a misunderstanding on my part on how systemd works with journald and rsyslog.  Nobody should take this thread seriously. #PEBKAC

PR #197, opened 2015-12-23
document new impstats format option

rgerhards:


radu-gheorghe:
Thanks for the option and the cod addition, @rgerhards!

PR #224, opened 2016-03-08
Add example output for cnum for non-numeric inputs.

dellis23:


rgerhards:
I am not sure if we want to document this in exactly that way. Did we guarantee somewhere that it behaves that way?

rgerhards:
I have thought a while, but I think we should not guarantee behaviour in those cases. This behaviour is not intended, just the way it currently happens to be.

PR #225, opened 2016-03-15
Fix typo in omfile failOnChOwnFailure description

mk-fg:


mk-fg:
Also, `Default: equally-named module parameter` description for the thing seem to be incorrect - there is no equally-named module parameter, neither in the docs, nor it is accepted by the code.
No idea what to replace it with though.

rgerhards:
Thanks, I have fixed the invalid description. The default is "on" (just checked the code).

PR #248, opened 2016-06-22
Added example for using multiple parsers within rulesets

crackytsi:
As there is no example within rsyslog documentation that gives an example how to use multiple Parsers in a ruleset I added this example.

### DIFF
--- a/source/configuration/parser.rst
+++ b/source/configuration/parser.rst
@@ -54,8 +54,8 @@ General Parser Parameters
    name is "pmciscoios", whereas it's default parser name is
    "rsyslog.pmciscoios".
 
-Sample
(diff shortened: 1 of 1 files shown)


crackytsi:
As it does not use the Parser object, I'm not sure if this is the best place.
http://www.rsyslog.com/doc/master/configuration/basic_structure.html#rulesets-and-rules could be a better place.
I expected a doucmentation page which covers all attributes of a ruleset.
Is this missing? 

regards
Chris

rgerhards:
I think the current position is good enough (we need to re-work a lot of things, anyhow). I think the ruleset object is indeed missing.

PR #281, opened 2016-11-24
WIP:overview.rst

mostolog:
First draft of documentation in markdown format.


mostolog:
Does RST allows embeding graphviz? I'll go for md, as de-facto standarized syntax.
Also, consider this would be a review process for all documentation, perhaps it's better to use an specific branch.

What do you use sphinx for?

rgerhards:
@mostolog  Please read the README files.
We cannot simply change to markdown, after all this is a community effort. If you want to take a different route, you need to go to the mailing list and convince all collaborators (or the majority) that your way is better ;-)

@radu-gheorghe can you help out in regard to graphviz?

rgerhards:
I was curios myself. Graphviz seems to be supported:
http://www.sphinx-doc.org/en/1.4.8/ext/graphviz.html
I do not exactly know which version we run on all involved systems, but I'd say give it a try.

mostolog:
I can squash commits, but I don't know how to fix CI (define a toc, add index/function/keywords)...could you give me a hand on this?

rgerhards:
A doc page that is never linked to will not be found and thus is worthless. So we do not permit this. Create a link:
http://www.sphinx-doc.org/en/1.4.9/markup/inline.html#ref-role

Sample:  imudp (https://raw.githubusercontent.com/rsyslog/rsyslog-doc/master/source/configuration/modules/imudp.rst) is linked e.g. from input description (https://raw.githubusercontent.com/rsyslog/rsyslog-doc/master/source/configuration/input.rst)

rgerhards:
and in case you haven't read the sphinx rest getting started guide, here it is: http://www.sphinx-doc.org/en/1.4.9/rest.html

rgerhards:
@mostolog have a look at #288 for a PR which adds a new page and also links to it from existing doc.

davidelang:
a few notes on the embedded graphs

graph shouation should be done locally, not pushed out to the Internet.

If possible, graphs should be generated from code (or at least linked documents) rather than being embedded in the doc directly

for example in overview.rst line 21 should not have to list every module, it should pull a list of modules of the various types (either from the source, or at least by looking at the im* and om* documents)

that same graph should also document mm* modules (being off to the side with arrows from and to the central bubble)

the central bubble should probably be called rsyslog rather than 'transform'

davidelang:
the graph on line 35 is incorrect, it shows:

im_ -> parser -> queue -> transform -> queue -> om_

the reality is simplified)

im* -> queue -> om*

or (more detailed)

im* -> queue -> worker thread


with the im* modules being responsible for populating the standard properties (some share configurable parser modules to do so)

the worker thread runs through the destination ruleset and does all the variable 'stuff', and applies all the filters, it then calls the code in all the output modules that are appropriate given the filters.

output modules almost always include transforming a message from the internal variables and properties to match the destination. Usually this is done by applying a template (which can be implemented as a sm* string module)

In the majority of configs out there, the worker thread doesn't actually set any variables or transform anything, it just applies filters and calls output module code.

davidelang:
line 51, rsyslog always uses rulesets, if you don't explicitly define one, it uses the default ruleset (i.e. anything that's not defined to be inside another ruleset)

davidelang:
line 56, the action() call is the output module in the normal case (it can also be a mm module)

any ruleset or action definition can also define a new queue, in which case the worker thread just delivers to that queue and there is a new set of worker threads to read from that queue

mbiebl:
I don't like this change, which uses images hosted/generated on another webserver.
I want to be able to read the documentation offline. Keep in mind that rsyslog-doc is shipped in distros as package for this specific reason. If this would require internet access, then having a package for it would be pointless.

The commit message also doesn't explain, why it would be a good idea to switch to such a solution.

Same for markdown vs rst: Please be super clear why such a switch would be beneficial, weighing its pros and cons. Every transition has its costs so should be well thought through.

davidelang:
if we are being pedantic (and this being the documentation, possibly we should be, but we need to layer the complexity). Rainer, please let me know if I'm missing something here

rsyslog has a housekeeping thread that manages other threads in two categories

1. input threads:
- created by input() statements and fetch/listen for new messages.
- populate the standard properties (may invoke a parser module)
- deliver the new messages to a queue (the main queue by default) 

2. worker threads process the ruleset associated with the queue[1] which can:
- set/clear a variable
- execute a filter (i.e. test) to decide if other statements are applicable to the log being processed
- call another ruleset [2], which executes and returns to continue procesing the current ruleset. Think of it as a subroutine call.
- call an output module (om*) [2], which takes the message variables/properties and delivers the message to something, almost always using a template [3] to control it's output
- call a message modification module (mm*) [4], which does something to the message variables/properties (adding, removing, renaming, changing any properties, including the default ones)
 

Where possible, input modules and worker threads don't work one message at a time, they work with a batch of messages, this allows a significant reduction in the amount of time that the queue is locked (one lock as the batch is started, one as it's finished rather than a pair of locks per message), and it allows the output modules to optimize their delivery by delivering a batch of messages at a time (link to batch documentation)


[1] qualifiers and implicit behavior of worker threads:
- there is a implicit ruleset on the main queue which contains everything not declared part of another ruleset
- one thread per queue by default, too many threads per queue decreases performance due to reducing batch sizes and lock congestion on the queues
- If the queue the worker is reading from is declared in an action() statement, there is an implicit ruleset containing that single action.

[2] if a ruleset or output module defines a queue, the worker thread just delivers to that queue

[3] a string module (sm*) is a C implementation of a template, it's called just like an interpreted template, but is drastically faster. 

[4] defining a queue on a mm* call is always an error as nothing else will see the results of the modules action. Some mm modules support using multiple threads to process messages in a batch.

davidelang:
Thank you for doing this documentation work. Reading this documentation and commenting on it is forcing me to think through things and in trying to explain them, forcing me to improve my grasp on what is happening (as the old saying goes, the best way to learn something is to teach it :-)

rgerhards:
2016-12-11 14:21 GMT+01:00 davidelang <notifications@github.com>:

> if we are being pedantic (and this being the documentation, possibly we
> should be, but we need to layer the complexity). Rainer, please let me know
> if I'm missing something here
>
> rsyslog has a housekeeping thread that manages other threads in two
> categories
>

actually, the housekeping that does not really "manage" the other threads.
What it does is:
- upon startup
   read config,  start threads
- one initialized
  do nothing except occasional janitor work
- when signalled to finish
  shut down other threads

>
>    1. input threads:
>
>
>    - created by input() statements and fetch/listen for new messages.
>    - populate the standard properties (may invoke a parser module)
>
> usually this is done at later stage, but depends on the input

>
>    - deliver the new messages to a queue (the main queue by default)
>
>
>    1. worker threads process the ruleset associated with the queue[1]
>    which can:
>
> parse message if not already done

>
>
>
>    - set/clear a variable
>    - execute a filter (i.e. test) to decide if other statements are
>    applicable to the log being processed
>    - call another ruleset [2], which executes and returns to continue
>    procesing the current ruleset. Think of it as a subroutine call.
>    - call an output module (om*) [2], which takes the message
>    variables/properties and delivers the message to something, almost always
>    using a template [3] to control it's output
>    - call a message modification module (mm*) [4], which does something
>    to the message variables/properties (adding, removing, renaming, changing
>    any properties, including the default ones)
>
> also template generation and other minor things that happen while a
message is being processed

>
>
> Where possible, input modules and worker threads don't work one message at
> a time, they work with a batch of messages, this allows a significant
> reduction in the amount of time that the queue is locked (one lock as the
> batch is started, one as it's finished rather than a pair of locks per
> message), and it allows the output modules to optimize their delivery by
> delivering a batch of messages at a time (link to batch documentation)
>
> [1] qualifiers and implicit behavior of worker threads:
>
>    - there is a implicit ruleset on the main queue which contains
>    everything not declared part of another ruleset
>    - one thread per queue by default, too many threads per queue
>    decreases performance due to reducing batch sizes and lock congestion on
>    the queues
>    - If the queue the worker is reading from is declared in an action()
>    statement, there is an implicit ruleset containing that single action.
>
> conceptually, you can say so. Actually it is a bit different (actions
existed long before we had the concept of multiple rulesets)

>
>
> [2] if a ruleset or output module defines a queue, the worker thread just
> delivers to that queue
>
> [3] a string module (sm*) is a C implementation of a template, it's called
> just like an interpreted template, but is drastically faster.
>
> [4] defining a queue on a mm* call is always an error
>

good point, config processor should probably better detect this ;-)

> as nothing else will see the results of the modules action. Some mm
> modules support using multiple threads to process messages in a batch.
>

Not sure if really such a mm* exists. However, this is true for all
modules, and inputs and outputs definitely do it (e.g. omfile with
background writing).

Rainer

davidelang:
a few clarifications.

When does the parser stack run?

A) in the input module
B) in the worker thread before anything else happens
C) something else

Is the template/string module called by the output module code to act on the message, or by the worker thread and the result passed to the output module?

re: rulesets/actions. I agree that actions existed before rulesets, but I think the explination I provided (with a default ruleset) is the clearest conceptually, and I don't think the history really affects the current reality.

I created issue https://github.com/rsyslog/rsyslog/issues/1323 for the mm action with queue case

rgerhards:
2016-12-12 8:46 GMT+01:00 davidelang <notifications@github.com>:

> a few clarifications.
>
> When does the parser stack run?
>
> A) in the input module
> B) in the worker thread before anything else happens
>

Mostly B, sometimes A, never C. IF A or B depends on the input module. E.g
imfile creates a fully populated message object, so it flags this as done
and the parser is not run again.

> C) something else
>
> Is the template/string module called by the output module code to act on
> the message, or by the worker thread and the result passed to the output
> module?
>

The worker thread *is* the thread the output modules runs on. So
thread-wise, there is not difference. Neverthless, the template is
generated by rsyslog core before calling into the output.

> re: rulesets/actions. I agree that actions existed before rulesets, but I
> think the explination I provided (with a default ruleset) is the clearest
> conceptually, and I don't think the history really affects the current
> reality.
>
Technically *it is* different because the queue workers have 0% in common.
It is still a valid picture to have on your mind when thinking about them,
but if you dig in and develop, it is pretty different.

> I created issue rsyslog/rsyslog#1323
> <https://github.com/rsyslog/rsyslog/issues/1323> for the mm action with
> queue case
>

Thanks,
Rainer

davidelang:
On Mon, 12 Dec 2016, Rainer Gerhards wrote:

> 2016-12-12 8:46 GMT+01:00 davidelang <notifications@github.com>:
>
>> When does the parser stack run?
>>
>> A) in the input module
>> B) in the worker thread before anything else happens
>>
>
> Mostly B, sometimes A, never C. IF A or B depends on the input module. E.g
> imfile creates a fully populated message object, so it flags this as done
> and the parser is not run again.

to make sure I have it straight.

An input module can do it's own parsing and fully populate all the standard 
properties.

If it does so, it sets a flag saying that parsing has succeeded.

When a worker thread grabs a message, if the flag is not set, the worker thread 
runs the parser stack (pm* modules).

So, for example, if we wanted to apply the parser stack to messages that imfile 
picks up, all that we would need to do is make the imfile module not set the 
flag.

>> Is the template/string module called by the output module code to act on
>> the message, or by the worker thread and the result passed to the output
>> module?
>>
>
> The worker thread *is* the thread the output modules runs on. So
> thread-wise, there is not difference. Neverthless, the template is
> generated by rsyslog core before calling into the output.

part of this is understanding what thread runs what code, part of this is 
looking at the call tree to understand what code is logically nested inside 
other code.

>> re: rulesets/actions. I agree that actions existed before rulesets, but I
>> think the explination I provided (with a default ruleset) is the clearest
>> conceptually, and I don't think the history really affects the current
>> reality.
>>
> Technically *it is* different because the queue workers have 0% in common.

This I don't understand.

The way I understand things, there is no difference between:

1. a worker that is working on the main queue using the default ruleset
2. a worker that is working a ruleset queue using the ruleset rules

other than what queue and rules they point at. Are you saying that the code for 
the two cases is different?

(I'm having a little truble visualizing exactly how an input that specifies a 
ruleset without a queue works. At that point it's still using the main queue, 
but a different ruleset. this seems like a problem, how does a worker thread 
know to grab the 'right' messages from the queue for the ruleset it's working? 
or does it figure out the ruleset to use for a particular message after it grabs 
a batch of messages off the main queue)

David Lang

rgerhards:
2016-12-12 9:22 GMT+01:00 davidelang <notifications@github.com>:

> On Mon, 12 Dec 2016, Rainer Gerhards wrote:
>
> > 2016-12-12 8:46 GMT+01:00 davidelang <notifications@github.com>:
> >
> >> When does the parser stack run?
> >>
> >> A) in the input module
> >> B) in the worker thread before anything else happens
> >>
> >
> > Mostly B, sometimes A, never C. IF A or B depends on the input module.
> E.g
> > imfile creates a fully populated message object, so it flags this as done
> > and the parser is not run again.
>
> to make sure I have it straight.
>
> An input module can do it's own parsing and fully populate all the
> standard
> properties.
>
>
or it can call the parser. There are a lot of subtle issues. IIRC , network
inputs need to call the parser when they are configured to do access
permissions based on hostname. The general rule is that an input should not
call parsers unless actually needed.


> If it does so, it sets a flag saying that parsing has succeeded.
>
>
yup


> When a worker thread grabs a message, if the flag is not set, the worker
> thread
> runs the parser stack (pm* modules).
>
>
yup


> So, for example, if we wanted to apply the parser stack to messages that
> imfile
> picks up, all that we would need to do is make the imfile module not set
> the
> flag.
>
>
which would also require changes to imfile itself, but basically "yes".


> >> Is the template/string module called by the output module code to act on
> >> the message, or by the worker thread and the result passed to the output
> >> module?
> >>
> >
> > The worker thread *is* the thread the output modules runs on. So
> > thread-wise, there is not difference. Neverthless, the template is
> > generated by rsyslog core before calling into the output.
>
> part of this is understanding what thread runs what code, part of this is
> looking at the call tree to understand what code is logically nested
> inside
> other code.
>
> >> re: rulesets/actions. I agree that actions existed before rulesets, but
> I
> >> think the explination I provided (with a default ruleset) is the
> clearest
> >> conceptually, and I don't think the history really affects the current
> >> reality.
> >>
> > Technically *it is* different because the queue workers have 0% in
> common.
>
> This I don't understand.
>
> The way I understand things, there is no difference between:
>
> 1. a worker that is working on the main queue using the default ruleset
> 2. a worker that is working a ruleset queue using the ruleset rules
>
> > other than what queue and rules they point at. Are you saying that the
code for

> the two cases is different?
>
>
that's the same, but you talked about action queues. They are different.



> (I'm having a little truble visualizing exactly how an input that
> specifies a
> ruleset without a queue works. At that point it's still using the main
> queue,
> but a different ruleset. this seems like a problem, how does a worker
> thread
> know to grab the 'right' messages from the queue for the ruleset it's
> working?
> or does it figure out the ruleset to use for a particular message after it
> grabs
> a batch of messages off the main queue)
>
>
The worker always grabs a batch from the queue. If the batch has different
rulesets associated, processing is somewhat slower. To be 100% preciese, I
need to re.read the code; I think it temporarily  generates new batches
consisting only of one ruleset.

For a rulesetworker, the worker process is runnig the ruleset logic.
For an action (Queue) worker, the worker process just calls the action
repeatedly (no filters, no logic, no nothing).

Rainer

mostolog:
Thanks for all your comments! I'll have a look on them and make changes :)

davidelang:
On Mon, 12 Dec 2016, Rainer Gerhards wrote:

> 2016-12-12 9:22 GMT+01:00 davidelang <notifications@github.com>:
>
>> On Mon, 12 Dec 2016, Rainer Gerhards wrote:
>>
>>> 2016-12-12 8:46 GMT+01:00 davidelang <notifications@github.com>:
>>>
>>>> When does the parser stack run?
>>>>
>>>> A) in the input module
>>>> B) in the worker thread before anything else happens
>>>>
>>>
>>> Mostly B, sometimes A, never C. IF A or B depends on the input module.
>> E.g
>>> imfile creates a fully populated message object, so it flags this as done
>>> and the parser is not run again.
>>
>> to make sure I have it straight.
>>
>> An input module can do it's own parsing and fully populate all the
>> standard
>> properties.
>>
>>
> or it can call the parser. There are a lot of subtle issues. IIRC , network
> inputs need to call the parser when they are configured to do access
> permissions based on hostname. The general rule is that an input should not
> call parsers unless actually needed.

would it simplify the code if the input module was responsible for doing the 
parsing/calling the parser and the worker could assume that it always received a 
parsed message?

Or is there concern that the parser could be slow enough to cause performance 
issues on the input side, so we prefer to do the parsing on the worker side of 
the queue where we can have multiple threads?

>> So, for example, if we wanted to apply the parser stack to messages that 
>> imfile picks up, all that we would need to do is make the imfile module not 
>> set the flag.
>>
>>
> which would also require changes to imfile itself, but basically "yes".

changes beyond imfile not setting the flag?

>>>> re: rulesets/actions. I agree that actions existed before rulesets, but I 
>>>> think the explination I provided (with a default ruleset) is the clearest 
>>>> conceptually, and I don't think the history really affects the current 
>>>> reality.
>>>>
>>> Technically *it is* different because the queue workers have 0% in common.
>>
>> This I don't understand.
>>
>> The way I understand things, there is no difference between:
>>
>> 1. a worker that is working on the main queue using the default ruleset
>> 2. a worker that is working a ruleset queue using the ruleset rules
>>
>>> other than what queue and rules they point at. Are you saying that the code 
>>> for the two cases is different?
>>
>>
> that's the same, but you talked about action queues. They are different.
>
>
>
>> (I'm having a little truble visualizing exactly how an input that specifies a 
>> ruleset without a queue works. At that point it's still using the main queue, 
>> but a different ruleset. this seems like a problem, how does a worker thread 
>> know to grab the 'right' messages from the queue for the ruleset it's 
>> working? or does it figure out the ruleset to use for a particular message 
>> after it grabs a batch of messages off the main queue)
>>
>>
> The worker always grabs a batch from the queue. If the batch has different
> rulesets associated, processing is somewhat slower. To be 100% preciese, I
> need to re.read the code; I think it temporarily  generates new batches
> consisting only of one ruleset.

so my take-away from this is that if you define a ruleset on an input, you 
really should have that ruleset use it's own queue. otherwise the batch 
processing suffers significantly if there are messages to be processed by 
multiple rulesets in the queue

worst case, if you had two rulesets sharing the main queue with perfect 
interleaving, the batch logic would lock the main queue once to grab the batch, 
but then make single-message batches for all processing.

I'm wondering if we should auto-create a queue for any rulesets referenced in an 
input() statement? I had to do too much 'black magic', but this seems like an 
easy pitfall to 'do the right (or at least a sane) thing' on to avoid

> For a rulesetworker, the worker process is runnig the ruleset logic.
> For an action (Queue) worker, the worker process just calls the action
> repeatedly (no filters, no logic, no nothing).

so a worker thread for the main queue or for a ruleset queue has logic to check 
for all types of statements, including calling the action code, but a worker 
thread for action queues is simplified to only run the template and action logic 
needed for that one action.

Is this something setup at startup time, or is it actually different code in the 
codebase (and would it simplifiy the codebase to refactor these to share code)?

David Lang

rgerhards:
upfront, so that I do not need to tell each time below: we are nearing a
level of detail where it would probably be better to review the actual
code. It is getting into very fine details ;-)

with that, on to the answers...

2016-12-12 11:31 GMT+01:00 davidelang <notifications@github.com>:

> On Mon, 12 Dec 2016, Rainer Gerhards wrote:
>
> > 2016-12-12 9:22 GMT+01:00 davidelang <notifications@github.com>:
> >
> >> On Mon, 12 Dec 2016, Rainer Gerhards wrote:
> >>
> >>> 2016-12-12 8:46 GMT+01:00 davidelang <notifications@github.com>:
> >>>
> >>>> When does the parser stack run?
> >>>>
> >>>> A) in the input module
> >>>> B) in the worker thread before anything else happens
> >>>>
> >>>
> >>> Mostly B, sometimes A, never C. IF A or B depends on the input module.
> >> E.g
> >>> imfile creates a fully populated message object, so it flags this as
> done
> >>> and the parser is not run again.
> >>
> >> to make sure I have it straight.
> >>
> >> An input module can do it's own parsing and fully populate all the
> >> standard
> >> properties.
> >>
> >>
> > or it can call the parser. There are a lot of subtle issues. IIRC ,
> network
> > inputs need to call the parser when they are configured to do access
> > permissions based on hostname. The general rule is that an input should
> not
> > call parsers unless actually needed.
>
> would it simplify the code if the input module was responsible for doing
> the
> parsing/calling the parser and the worker could assume that it always
> received a
> parsed message?
>

No, not notable. We had this in the past ...

>
> Or is there concern that the parser could be slow enough to cause
> performance
> issues on the input side, so we prefer to do the parsing on the worker
> side of
> the queue where we can have multiple threads?


and intentionally moved it out for the reason you give. The performance
impact was pretty noticable. I *think* there are also cases where some
properties are never evaluated if not used inside template. But that's
probably more for fromhost & fromhost-ip (code review needed for 100%
verfication)



>
> >> So, for example, if we wanted to apply the parser stack to messages
> that
> >> imfile picks up, all that we would need to do is make the imfile module
> not
> >> set the flag.
> >>
> >>
> > which would also require changes to imfile itself, but basically "yes".
>
> changes beyond imfile not setting the flag?


It would also need to not construct the message object. Right now, it sets
tags, hostname, etc. This would need to be disabled in thoses case.


>
>
> >>>> re: rulesets/actions. I agree that actions existed before rulesets,
> but I
> >>>> think the explination I provided (with a default ruleset) is the
> clearest
> >>>> conceptually, and I don't think the history really affects the
> current
> >>>> reality.
> >>>>
> >>> Technically *it is* different because the queue workers have 0% in
> common.
> >>
> >> This I don't understand.
> >>
> >> The way I understand things, there is no difference between:
> >>
> >> 1. a worker that is working on the main queue using the default ruleset
> >> 2. a worker that is working a ruleset queue using the ruleset rules
> >>
> >>> other than what queue and rules they point at. Are you saying that the
> code
> >>> for the two cases is different?
> >>
> >>
> > that's the same, but you talked about action queues. They are different.
> >
> >
> >
> >> (I'm having a little truble visualizing exactly how an input that
> specifies a
> >> ruleset without a queue works. At that point it's still using the main
> queue,
> >> but a different ruleset. this seems like a problem, how does a worker
> thread
> >> know to grab the 'right' messages from the queue for the ruleset it's
> >> working? or does it figure out the ruleset to use for a particular
> message
> >> after it grabs a batch of messages off the main queue)
> >>
> >>
> > The worker always grabs a batch from the queue. If the batch has
> different
> > rulesets associated, processing is somewhat slower. To be 100% preciese,
> I
> > need to re.read the code; I think it temporarily generates new batches
> > consisting only of one ruleset.
>
> so my take-away from this is that if you define a ruleset on an input, you
> really should have that ruleset use it's own queue. otherwise the batch
> processing suffers significantly if there are messages to be processed by
> multiple rulesets in the queue
>

yes, this was the recommendation from day one. For a low-end system, it's
not needed, though.


>
> worst case, if you had two rulesets sharing the main queue with perfect
> interleaving, the batch logic would lock the main queue once to grab the
> batch,
> but then make single-message batches for all processing.
>
>
I *think* (code review required) that in this case two sub-batches are
build and submitted. So there is memory copy overhead, but not the
single-message processing.


> I'm wondering if we should auto-create a queue for any rulesets referenced
> in an
> input() statement? I had to do too much 'black magic', but this seems like
> an
> easy pitfall to 'do the right (or at least a sane) thing' on to avoid


Dangerous, folks may use the same ruleset for a sync subrouotine call. We
could check for that as well, but with the new call_indirect, we can no
longer do this. It would also require time to do it. But with
call_indirect, that's really not an option.


>
>
> > For a rulesetworker, the worker process is runnig the ruleset logic.
> > For an action (Queue) worker, the worker process just calls the action
> > repeatedly (no filters, no logic, no nothing).
>
> so a worker thread for the main queue or for a ruleset queue has logic to
> check
> for all types of statements, including calling the action code, but a
> worker
> thread for action queues is simplified to only run the template and action
> logic
> needed for that one action.
>
> Is this something setup at startup time, or is it actually different code
> in the
> codebase (and would it simplifiy the codebase to refactor these to share
> code)?
>
>
no, its really two different functions. The queue code is generic, though.
It has a pointer to the queue consumer (being either the ruleset code or
the simple action code).

On generalization: in theory it is good to have everything abstracted. If
you go into tuning and algo engineering, that's not always the case. For
example, we have (review required for exact number) 4 or 5 different
functions/code path for enqueing a message. To narrow the sample, we have
one function that handles cases where repeated message reduction is on and
one for cases where repeated message reduction is off. There is lot of code
duplicated, so one must be careful when maintaining this. It's not even
possible to abstract this aways with macros. HOWEVER, this duplicate code
is much faster (10+%) than a generalized single body of code, which *just
checks if repeated message reduction is set* and acts accordingly. The
*checks alone* make the difference here. I know for sure, because we
started with a single function, and (in v7?) removed the abstraction and
duplicated code, resulting in a big processing gain.

There are many micro-optimizations in rsyslog, this is why I say if we dig
deeper, we need to go and read source code because it becomes to complex to
describe otherwise.

HTH
Rainer

davidelang:
On Mon, 12 Dec 2016, Rainer Gerhards wrote:

> 2016-12-12 11:31 GMT+01:00 davidelang <notifications@github.com>:
>
>> Or is there concern that the parser could be slow enough to cause performance 
>> issues on the input side, so we prefer to do the parsing on the worker side 
>> of the queue where we can have multiple threads?
>
> and intentionally moved it out for the reason you give. The performance
> impact was pretty noticable. I *think* there are also cases where some
> properties are never evaluated if not used inside template. But that's
> probably more for fromhost & fromhost-ip (code review needed for 100%
> verfication)

wow, that should also mean that disabling DNS lookups should have no effect 
unless you actually use fromhost. nice.

>>>> So, for example, if we wanted to apply the parser stack to messages that 
>>>> imfile picks up, all that we would need to do is make the imfile module not 
>>>> set the flag.
>>>>
>>>>
>>> which would also require changes to imfile itself, but basically "yes".
>>
>> changes beyond imfile not setting the flag?
>
>
> It would also need to not construct the message object. Right now, it sets
> tags, hostname, etc. This would need to be disabled in thoses case.

Ok, I was thinking we could get away with letting imfile set them and the parser 
overwrite them.

>> so my take-away from this is that if you define a ruleset on an input, you
>> really should have that ruleset use it's own queue. otherwise the batch
>> processing suffers significantly if there are messages to be processed by
>> multiple rulesets in the queue
>>
>
> yes, this was the recommendation from day one. For a low-end system, it's
> not needed, though.

I don't remember seeing this anywhere in the docs.

>> worst case, if you had two rulesets sharing the main queue with perfect 
>> interleaving, the batch logic would lock the main queue once to grab the 
>> batch, but then make single-message batches for all processing.
>>
>>
> I *think* (code review required) that in this case two sub-batches are
> build and submitted. So there is memory copy overhead, but not the
> single-message processing.
>
>
>> I'm wondering if we should auto-create a queue for any rulesets referenced in 
>> an input() statement? I had to do too much 'black magic', but this seems like 
>> an easy pitfall to 'do the right (or at least a sane) thing' on to avoid
>
>
> Dangerous, folks may use the same ruleset for a sync subrouotine call. We
> could check for that as well, but with the new call_indirect, we can no
> longer do this. It would also require time to do it. But with
> call_indirect, that's really not an option.

I was thinking of only doing this for rulesets used in input statements, but 
you're right, someone could use a ruleset both for that and as a sync subroutine 
from some other ruleset. I guess a warning is all we can do.

David Lang

mostolog:
After spending a few hours reading and trying to understand all your comments (To be honest, at the end, when it became TOO deep for overview.rst, I skipped a few lines), here are some comments I'll love to give you before editing overview.rst

@davidelang 
> Thank you for doing this documentation work. Reading this documentation and commenting on it is forcing me to think through things and in trying to explain them, forcing me to improve my grasp on what is happening (as the old saying goes, the best way to learn something is to teach it :-)

That's why I started writing it. You're always welcome


> graph shouation should be done locally, not pushed out to the Internet.

Agree. But I wasn't able to do it on github. I'll try to find another way.

> If possible, graphs should be generated from code (or at least linked documents) rather than being embedded in the doc directly

I disagree. Having dot syntax within documentation make it easier to mantain/changes, rather than _change, compile and link new images_

> for example in overview.rst line 21 should not have to list every module, it should pull a list of modules of the various types (either from the source, or at least by looking at the im* and om* documents)

If it were so easy...I thought _dot within rst_ was the best option. Doing so I think it can't be done on github.

> that same graph should also document mm* modules (being off to the side with arrows from and to the central bubble)

Advanded knowledge required. I'll update that.


> the worker thread runs through the destination ruleset and does all the variable 'stuff', and applies all the filters, it then calls the code in all the output modules that are appropriate given the filters.

> In the majority of configs out there, the worker thread doesn't actually set any variables or transform anything, it just applies filters and calls output module code.

> Rainer: execute a filter (i.e. test) to decide if other statements are applicable to the log being processed

please, elaborate on "filters".


> output modules almost always include transforming a message from the internal variables and properties to match the destination. Usually this is done by applying a template (which can be implemented as a sm* string module)

sm modules are "expert" features I don't think should appear on overview. Thoughts?


@mbiebl
> I don't like this change, which uses images hosted/generated on another webserver.
>
> I want to be able to read the documentation offline. Keep in mind that rsyslog-doc is shipped in distros as package for this specific reason. If this would require internet access, then having a package for it would be pointless.
>
> The commit message also doesn't explain, why it would be a good idea to switch to such a solution.

Totally agree. Yet to find a way to have offline embebed in rst graphics.

> Same for markdown vs rst: Please be super clear why such a switch would be beneficial, weighing its pros and cons. Every transition has its costs so should be well thought through.

rst is the official documentation format. Are there any other .md documents apart from README.md?

@davidelang 
> part of this is understanding what thread runs what code, part of this is looking at the call tree to understand what code is logically nested inside other code.

sequence diagrams to the rescue


>[4] defining a queue on a mm* call is always an error as nothing else will see the results of the modules action. Some mm modules support using multiple threads to process messages in a batch.

>>> Not sure if really such a mm* exists. However, this is true for all modules, and inputs and outputs definitely do it (e.g. omfile with background writing).

Let me see if I understood properly. mm* read from queue, do some magic and return the messages to the queue. If a queue is defined within action, the message will be stored in lala-land, where nobody read it. right?


> I think it temporarily generates new batches consisting only of one ruleset. For a rulesetworker, the worker process is runnig the ruleset logic. For an action (Queue) worker, the worker process just calls the action repeatedly (no filters, no logic, no nothing).

Let me see if I understood: using "main ruleset(no ruleset)", workers just start working, but if using ruleset without a queue, they need to create an in-memory queue, hence having worse performance. AKA: whenever you use a ruleset, specify a queue!!!

Please, remember that's overview.rst not hardcode-internals-for-damn-pros.rst

davidelang:
On Wed, 14 Dec 2016, mostolog wrote:

>> the worker thread runs through the destination ruleset and does all the variable 'stuff', and applies all the filters, it then calls the code in all the output modules that are appropriate given the filters.
>
>> In the majority of configs out there, the worker thread doesn't actually set any variables or transform anything, it just applies filters and calls output module code.
>
>> Rainer: execute a filter (i.e. test) to decide if other statements are applicable to the log being processed
>
> please, elaborate on "filters".

if ... then

*.info

:msg, contains, "foo"

are all filters/tests

>> output modules almost always include transforming a message from the internal variables and properties to match the destination. Usually this is done by applying a template (which can be implemented as a sm* string module)
>
> sm modules are "expert" features I don't think should appear on overview. Thoughts?

probably not (other than possibly a footnote when talking about templates)


> @davidelang
>> part of this is understanding what thread runs what code, part of this is 
>> looking at the call tree to understand what code is logically nested inside 
>> other code.
>
> sequence diagrams to the rescue
>
>
>> [4] defining a queue on a mm* call is always an error as nothing else will 
>> see the results of the modules action. Some mm modules support using multiple 
>> threads to process messages in a batch.
>
>>>> Not sure if really such a mm* exists. However, this is true for all 
>>>> modules, and inputs and outputs definitely do it (e.g. omfile with 
>>>> background writing).
>
> Let me see if I understood properly. mm* read from queue, do some magic and 
> return the messages to the queue. If a queue is defined within action, the 
> message will be stored in lala-land, where nobody read it. right?

conceptually yes, but the message doesn't actually get stored backon the queue.

without an action queue on the mm modules it is:

grab a batch of messages from the queue.

foreach message (
   do stuff
   execute mm module which modifies the message object the thread is working on
   do more stuff
   finished working on this copy of the log message
)

with an action queue it's:

grab a batch of messages from the queue.

foreach message (
   do stuff
   put a copy of the message on the mm module queue
   do more stuff
   finished working on this copy of the log message
)


in a different thread (worker thread for the mm module queue

grab a batch of messages from the mm module action queue

foreach message (
   execute mm module which modifies the message object that this worker thread is 
working on (the copy from the action queue)
   finished working on this copy of the log message
)

there is no 'do more stuff' step possible, so the changes the mm module makes 
are thrown away immediately after.

>> I think it temporarily generates new batches consisting only of one ruleset. 
>> For a rulesetworker, the worker process is runnig the ruleset logic. For an 
>> action (Queue) worker, the worker process just calls the action repeatedly 
>> (no filters, no logic, no nothing).
>
> Let me see if I understood: using "main ruleset(no ruleset)", workers just 
> start working, but if using ruleset without a queue, they need to create an 
> in-memory queue, hence having worse performance. AKA: whenever you use a 
> ruleset, specify a queue!!!

not quite. if you have a ruleset without a queue (specified in an input 
statement, other rulesets don't have this issue), then it's not that an 
additional queue is created, but rather that the batch gets broken down into 
smaller batches to be processed. Batches were created because it can be FAR more 
efficient to process lots of messages at once.

The case that I had in mind when we created batches was inserting messages into 
a database, where the cost of doing two inserts of one message each can be about 
the same as the cost of creating one insert of 100-1000 messages.

We also found that the cost of the internal locking was a very large percentage 
of the cost of processing a message, so moving to large batches gave us a 
10-100x speedup just due to the reduced lock contention inside rsyslog.

> Please, remember that's overview.rst not hardcode-internals-for-damn-pros.rst

correct, the reason for going into depth like I did was to make sure that you 
(and I) understand exactly what's happening so that we can describe it properly 
(and simplify it appropriately)

the functional diagrams you initally created showed a misunderstanding of how 
the internals work (and my initial corrections also showed some 
misunderstandings on my end, which is why we went down into the weeds to 
understand exactly how things do work)

David Lang

mostolog:
So..."worker grabs a few messages from queue and make changes"?

    action(type="mm...without queue...)

while, "worker gets a copy of the message and make changes, which are lost cause it's a copy"?

    action(type="mm...with queue...)

Is that correct?

Also, this is slower:

    input(ruleset="foo")
    ruleset(...without queue...)

than this:

    input(ruleset="foo")
    ruleset(...with queue...)

Am I right?

rgerhards:
>
> graph shouation should be done locally, not pushed out to the Internet.
>
> Agree. But I wasn't able to do it on github. I'll try to find another way.
>

I posted a link early in the thread (when you asked the same question) and
it tells you how to do it properly with RST and sphinx. I suggest you read
it and try if it works for you ;-)

...


> for example in overview.rst line 21 should not have to list every module,
> it should pull a list of modules of the various types (either from the
> source, or at least by looking at the im* and om* documents)
>
> If it were so easy...I thought *dot within rst* was the best option.
> Doing so I think it can't be done on github.
>

just to be cristal-clear here: github hosts the **source** repository for
the documentation. Other than that, github has NOTHING to do with the doc.

The link I provided states the official sphinx way of doing dot in RST.

that same graph should also document mm* modules (being off to the side
> with arrows from and to the central bubble)
>
> Advanded knowledge required. I'll update that.
>
> the worker thread runs through the destination ruleset and does all the
> variable 'stuff', and applies all the filters, it then calls the code in
> all the output modules that are appropriate given the filters.
>
> In the majority of configs out there, the worker thread doesn't actually
> set any variables or transform anything, it just applies filters and calls
> output module code.
>
> Rainer: execute a filter (i.e. test) to decide if other statements are
> applicable to the log being processed
>
> please, elaborate on "filters".
>

http://www.rsyslog.com/doc/v8-stable/configuration/filters.html

> output modules almost always include transforming a message from the
> internal variables and properties to match the destination. Usually this is
> done by applying a template (which can be implemented as a sm* string
> module)
>
> sm modules are "expert" features I don't think should appear on overview.
> Thoughts?
>
> @mbiebl <https://github.com/mbiebl>
>
> I don't like this change, which uses images hosted/generated on another
> webserver.
>
> I want to be able to read the documentation offline. Keep in mind that
> rsyslog-doc is shipped in distros as package for this specific reason. If
> this would require internet access, then having a package for it would be
> pointless.
>
> The commit message also doesn't explain, why it would be a good idea to
> switch to such a solution.
>
> Totally agree. Yet to find a way to have offline embebed in rst graphics.
>
> Same for markdown vs rst: Please be super clear why such a switch would be
> beneficial, weighing its pros and cons. Every transition has its costs so
> should be well thought through.
>
> rst is the official documentation format. Are there any other .md
> documents apart from README.md?
>
NO


> @davidelang <https://github.com/davidelang>
>
> part of this is understanding what thread runs what code, part of this is
> looking at the call tree to understand what code is logically nested inside
> other code.
>
> sequence diagrams to the rescue
>
> [4] defining a queue on a mm* call is always an error as nothing else will
> see the results of the modules action. Some mm modules support using
> multiple threads to process messages in a batch.
>
> Not sure if really such a mm* exists. However, this is true for all
> modules, and inputs and outputs definitely do it (e.g. omfile with
> background writing).
>
> Let me see if I understood properly. mm* read from queue, do some magic
> and return the messages to the queue. If a queue is defined within action,
> the message will be stored in lala-land, where nobody read it. right?
>

http://www.rsyslog.com/doc/queues_analogy.html

modified in action queue, no other action, discarded


> I think it temporarily generates new batches consisting only of one
> ruleset. For a rulesetworker, the worker process is runnig the ruleset
> logic. For an action (Queue) worker, the worker process just calls the
> action repeatedly (no filters, no logic, no nothing).
>
> Let me see if I understood: using "main ruleset(no ruleset)", workers just
> start working, but if using ruleset without a queue, they need to create an
> in-memory queue, hence having worse performance. AKA: whenever you use a
> ruleset, specify a queue!!!
>

old, but still valid:

http://www.rsyslog.com/doc/v8-stable/concepts/queues.html

Rainer

davidelang:
On Wed, 14 Dec 2016, mostolog wrote:

> So..."worker grabs a few messages from queue and make changes"?
>
>    action(type="mm...without queue...)
>
> while, "worker gets a copy of the message and make changes, which are lost cause it's a copy"?
>
>    action(type="mm...with queue...)
>
> Is that correct?

yes

> Also, this is slower:
>
>    input(ruleset="foo")
>    ruleset(...without queue...)
>
> than this:
>
>    input(ruleset="foo")
>    ruleset(...with queue...)
>
> Am I right?

yes

mostolog:
I have updated overview to reflect some changes you mention.
Consider it still uses "online graphs" for easier working. neither I have properly linked the page, but I'll love to get some content feedback.

Thanks

README.rst
```diff
@@ -1,3 +1,5 @@
+.. index:: ! readme
+
 Rsyslog-doc
 ===========
 This is the Documentation for `rsyslog <https://github.com/rsyslog/rsyslog>`_ project.
```
rgerhards: This needs to link to the official home page, which is www.rsyslog.com.

mostolog:
@rgerhards Could you fix CI issue:
_/home/travis/build/rsyslog/rsyslog-doc/source/overview.rst:15: ERROR: Unknown directive type "graphviz"._ ?

PR #287, opened 2016-12-09
mmrm1stspace: add documentation

PascalWithopf:


source/configuration/modules/mmrm1stspace.rst
```diff
@@ -0,0 +1,27 @@
...
+
+This example receives messages over imtcp and modifies them, before sending
+them to a file.
+
+::
+   module(load="imtcp")
+   module(load="mmrm1stspace")
+   input(type="imtcp" port="13514")
+   action(type="mmrm1stspace")
+   action(type="omfile" template="outfmt" file="rsyslog.out.log")
```
rgerhards: The template used here is not defined, so this does not work if someone tries it out. As the template itself is not important, I suggest to remove it. Also, a somewhat more common output file name may look nicer (e.g. "output.log") - but that's pesonal taste.

PR #292, opened 2016-12-19
WIP: Imfile

mostolog:
**Updated imfile doc, feedback requested**

Still, needing some more info about:

* On linux, files are monitored using inodes, which enable file renaming without having path issues. How this is handled on Windows?
* How inotify mode work internally on Linux? and on Windows?
* Accoding to doc, theres a Rainer blog article about escapeLF. Is this one?: http://blog.gerhards.net/2013/09/imfile-multi-line-messages.html

### DIFF
--- a/source/configuration/modules/imfile.rst
+++ b/source/configuration/modules/imfile.rst
@@ -38,7 +38,7 @@ Monitoring for changes supports file rotation under some circunstances:
 * rsyslog must be running while file is rotated, otherwise lines after last reading position won't be processed
 * once rsyslog ends with remaining lines, it will start reading new file from the beginning.
 
-Have in mind that files aren't monitored using paths, but have a *file handler* which doesn't change if file is renamed.
+Keep in mind that files aren't monitored using paths, but have a *file handler* which doesn't change if file is renamed.
 
 Module Parameters
 *****************
--- a/source/configuration/modules/imfile.rst
+++ b/source/configuration/modules/imfile.rst
@@ -40,8 +40,6 @@ Monitoring for changes supports file rotation under some circunstances:
 
 Have in mind that files aren't monitored using paths, but have a *file handler* which doesn't change if file is renamed.
 
-**TODO is that true for Windows also?**
-
 Module Parameters
 *****************
 
@@ -54,13 +52,9 @@ Available since:  8.1.5
 
 This specifies if imfile is shall run in *inotify* or *polling* mode.
 
-inotify:
+inotify: it works by registering with the OS to be notified when the file changes.
 
-**TODO add a description about how inotify works internally**
-
-Although lines are read *sooner* than using *polling* mode, there may be a slight delay due to internal processing.
-
-polling: it needs more resources and its slower than *inotify*, reason why **mode="inotify" is recommended**.
+polling: it needs more resources and its slower than *inotify*, reason why **mode="inotify" is recommended** if supported by your system.
 Basically there's a timeout at which file is checked for changes.
 
 ================  ======================
@@ -84,7 +78,7 @@ Default:      none
 Legacy:       $InputFileName
 ============  ===============
 
-File(s) being processed. This must be an absolute file path (no macros or templates). 
+File(s) being processed. This shall be an absolute file path (no macros or templates). 
 
 This option currently supports *file wilcards* in **inotify mode only** (it hasn't been implemented for polling) for monitoring multiple files. Consider wildcards are only valid at directory level, ie:
 
@@ -131,6 +125,7 @@ Default:  none
 Legacy:   $InputFileBindRuleset
 ========  =====================
 
+**TODO link ruleset**
 Binds the listener to a specific Ruleset.
 
 ============  =============================
@@ -194,13 +189,15 @@ Provides support for *simple* multi-line processing. Incompatible with *startmsg
 * 1 - paragraph: there is a blank line between log messages.
 * 2 - indented: new messages start at the beginning of a line. If a line starts with a space or tab ("\\t") it is part of the previous message.
 
+**This paramater is deprecated. DO NOT use it, because it would be removed in future releases**.
+
 ================  =================
 startmsg.regex:   "POSIX ERE regex"
 Default:          none
 Available since:  8.10.0
 ================  =================
 
-This option allows more complex multi-line messages, using provided regex as a start of new messages. As it is using regular expressions, it's more flexible than *readMode* but at the cost of lower performance.
+This option allows more complex multi-line messages, using provided regex as a start of new messages. As it is using regular expressions, it's more flexible than *readMode* but at the cost of lower performance when working with complex regular expressions.
 
 Note that *readMode* and *startmsg.regex* cannot both be defined for the same input.
 
--- a/source/configuration/modules/imfile.rst
+++ b/source/configuration/modules/imfile.rst
@@ -1,570 +1,408 @@
+.. index:: ! imfile
+
 imfile: Text File Input Module
-==============================
+##############################
 
-.. index:: ! imfile 
+================  ===========================================================================
+**Module Name:**  **imfile**
+**Author:**       `Rainer Gerhards <http://www.gerhards.net/rainer>`_ <rgerhards@adiscon.com>
+================  ===========================================================================
 
-===========================  ===========================================================================
-**Module Name:**             **imfile**
-**Author:**                  `Rainer Gerhards <http://www.gerhards.net/rainer>`_ <rgerhards@adiscon.com>
-===========================  ===========================================================================
+Use this module to read **text files** to process them on rsyslog.
+In a text file:
 
-This module provides the ability to convert any standard text file
-into a syslog
-message. A standard text file is a file consisting of printable
-characters with lines being delimited by LF.
+* lines have printable characters
+* lines are delimited by *LF*
 
-The file is read line-by-line and any line read is passed to rsyslog's
-rule engine. The rule engine applies filter conditions and selects which
-actions needs to be carried out. Empty lines are **not** processed, as
-they would result in empty syslog records. They are simply ignored.
+Please, notice we talk about **reading text lines, not syslog messages stored on a file**, therefore it makes sense to have parameters to set *tag, priority or severity*.
 
-As new lines are written they are taken from the file and processed.
-Depending on the selected mode, this happens via inotify or based on
-a polling interval. Especially in polling mode, file reading doesn't
-happen immediately. But there are also slight delays (due to process
-scheduling and internal processing) in inotify mode.
+Files are read line-by-line **ignoring any empty lines**, and rsyslog keeps information about *last reading position* in what's called **state files**.
 
-The file monitor supports file rotation. To fully work,
-rsyslogd must run while the file is rotated. Then, any remaining lines
-from the old file are read and processed and when done with that, the
-new file is being processed from the beginning. If rsyslogd is stopped
-during rotation, the new file is read, but any not-yet-reported lines
-from the previous file can no longer be obtained.
+State files
+***********
 
-When rsyslogd is stopped while monitoring a text file, it records the
-last processed location and continues to work from there upon restart.
-So no data is lost during a restart (except, as noted above, if the file
-is rotated just in this very moment).
+State files are used to store the last reading position of a file and they are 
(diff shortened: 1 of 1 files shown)


mostolog:
apart from squashing...any comments?

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
-   knows exactly why this is required.
+This specifies if imfile is shall run in *inotify* or *polling* mode.
 
-.. index:: 
-   single: imfile;  Ruleset
-.. function:: Ruleset <ruleset> 
+inotify:
 
-   Binds the listener to a specific :doc:`ruleset <../../concepts/multi_ruleset>`.
+**TODO add a description about how inotify works internally**
```
davidelang: Polling works by waking up every X seconds and checking for new data in the file. Inotify works by registering with the OS to be notified when the file changes.

Inotify is less load on the system than polling, but it does require more support from the kernel and filesystem, so there can be cases where polling is desired

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
 
-.. function:: addMetadata [on/off]
+Although lines are read *sooner* than using *polling* mode, there may be a slight delay due to internal processing.
```
davidelang: please remove this. It adds more confusion than information.

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
 
-   **Default: see intro section on Metadata**
+polling: it needs more resources and its slower than *inotify*, reason why **mode="inotify" is recommended**.
+Basically there's a timeout at which file is checked for changes.
 
-   This is used to turn on or off the addition of metadata to the
-   message object.
+================  ======================
```
davidelang: also move to expert section along with the inotify/polling option

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
-
-   This is an **experimental** feature that tells rsyslog to reopen input file
-   when it was truncated (inode unchanged but file size on disk is less than
-   current offset in memory).
-
-.. index::
-   single: imfile; trimLineOverBytes
-.. function:: trimLineOverBytes [number] (requires v8.17.0+)
+============  ===============
+**REQUIRED**
```
davidelang: As a meta-comment, it's probably best to put the required options first, then the optional, then the expert (although optional items that are tied closely with a required one may belong right under the required option)

the idea is to make it as trivial as possible for someone to configure the simple (hopefully most common) use case, and then dig down into the details when they need to.

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
+file:         "/path/to/file"
+Default:      none
+Legacy:       $InputFileName
+============  ===============
 
-   **Default: 0**
+File(s) being processed. This must be an absolute file path (no macros or templates).
```
davidelang: FYI, I think it can be a relative path to where rsyslog is started, but it's a _really_ bad idea to depend on that.

so I'd leave the comment as-is in the documentation, but it's good for you to know this.

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
-.. function:: freshStartTail [on/off] (requires v8.18.0+)
+============  =============
+**REQUIRED**
+tag:          "value"
+Default:      none
+Legacy:       $InputFileTag
+============  =============
 
-   **Default: off**
+**TODO: preperly link Syslog Specification https://tools.ietf.org/html/rfc5424#section-6.2.5**
```
davidelang: just put this link in the doc. People almost never need to reference it, so I think we are good just putting a link here, even for people reading this without a network connection.

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
+ 
+=========  ==================
+facility:  "value"
+Default:   "local0"
+Legacy:    $InputFileFacility
+=========  ==================
 
-Caveats/Known Bugs
-------------------
+**TODO: preperly link Syslog Specification https://tools.ietf.org/html/rfc5424#section-6.2.5**
```
davidelang: same as above

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
+Legacy:    $InputFileSeverity
+=========  ==================
 
-The following sample monitors two files. If you need just one, remove
-the second one. If you need more, add them according to the sample ;).
-This code must be placed in /etc/rsyslog.conf (or wherever your distro
-puts rsyslog's config files). Note that only commands actually needed
-need to be specified. The second file uses less commands and uses
-defaults instead.
+**TODO: preperly link Syslog Specification https://tools.ietf.org/html/rfc5424#section-6.2.5**
```
davidelang: same

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
 
-::
+The syslog severity to be assigned to lines read. Textual form is suggested (e.g. "info", "warning", ...) but numbers are also accepted (e.g. 4 for "info") as defined in RFC.
 
-  module(load="imfile" PollingInterval="10") #needs to be done just once 
+========  =====================
```
davidelang: add a link here to the ruleset documentation.

This is borderline expert mode.

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
-Note: in order to preserve compatibility with previous versions, the LF escaping
-in multi-line messages is turned off for legacy-configured file monitors
-(the "escapeLF" input parameter). This can cause serious problems. So it is highly
-suggested that new deployments use the new :ref:`input() <cfgobj_input>` configuration 
-object and keep LF escaping turned on. 
+By default, it's set to *on* if *file* contains *wildcards*, otherwise default is *off*.
 
-Legacy Configuration Directives
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+==================  ====================
```
davidelang: advanced option, should be listed after readmode

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
+Default:   "0"
+=========  =======
 
-   equivalent to: "Severity"
+Provides support for *simple* multi-line processing. Incompatible with *startmsg.regex*. 
 
-.. index:: 
-   single: imfile; $InputRunFileMonitor
-.. function:: $InputRunFileMonitor
+* 0 - line: each line is a new message.
```
davidelang: with the addition of the startmsg.regex, it's probably a good idea to mark this as depriciated and list what the appropriate regex would be to use instead.

In the future, the internal implementation of these may change to do this anyway.

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
-   forget this directive, no file monitoring will take place.
+================  =================
+startmsg.regex:   "POSIX ERE regex"
+Default:          none
+Available since:  8.10.0
+================  =================
 
-.. index:: 
-   single: imfile; $InputFilePollInterval
+This option allows more complex multi-line messages, using provided regex as a start of new messages. As it is using regular expressions, it's more flexible than *readMode* but at the cost of lower performance.
```
davidelang: per Rainer, this is not a significant performance impact if the regex is simple. we should test this with the readmode-equivalent regexes and if it really is the same (or very close) change the performance comment to be that complex regexes can be expensive rather than saying that this is slower than readmode

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
+Specifying a value of zero turns off timeout processing. Also note the
+relationship to the *timeoutGranularity* parameter, which sets the
+lower bound of *readTimeout*.
+
+Setting timeout vaues slightly increases processing time requirements; the
+effect should only be visible of a very large number of files is being
+monitored.
+
+===================  ================
+timeoutgranularity:  "seconds"
```
davidelang: mark as expert mode, but leave it here

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
+drawbacks and it's now deprecated.
+
+Monitoring for changes supports file rotation under some circunstances:
+
+* rsyslog must be running while file is rotated, otherwise lines after last reading position won't be processed
+* once rsyslog ends with remaining lines, it will start reading new file from the beginning.
+
+Have in mind that files aren't monitored using paths, but have a *file handler* which doesn't change if file is renamed.
+
+**TODO is that true for Windows also?**
```
mostolog: Ok.

source/configuration/modules/imfile.rst
```diff
@@ -1,570 +1,408 @@
...
-   single: imfile; MaxSubmitAtOnce
-.. function:: MaxSubmitAtOnce [number]
-
-   This is an expert option. It can be used to set the maximum input
-   batch size that imfile can generate. The **default** is 1024, which
-   is suitable for a wide range of applications. Be sure to understand
-   rsyslog message batch processing before you modify this option. If
-   you do not know what this doc here talks about, this is a good
-   indication that you should NOT modify the default.
+*****************
```
mostolog: Just applied top-bottom: http://www.sphinx-doc.org/en/1.5.1/rest.html#sections
    # with overline, for parts
    * with overline, for chapters
    =, for sections
    -, for subsections
    ^, for subsubsections
    ", for paragraphs

I don't mind to change to anything once you and @rgerhards agree a common format/style.

source/configuration/modules/imfile.rst
mostolog: Ok.

source/configuration/modules/imfile.rst
mostolog: Ok.

source/configuration/modules/imfile.rst
mostolog: Do you mean creating a header for each: required/optional/multinie/expert/experimental?

source/configuration/modules/imfile.rst
mostolog: shall ;)

source/configuration/modules/imfile.rst
mostolog: Just had some issues with rst syntax. As link is referenced in multiple places. still had to figure it out how to have the link "several times" within the same page.

davidelang:
On Thu, 12 Jan 2017, mostolog wrote:

> +Have in mind that files aren't monitored using paths, but have a *file handler* which doesn't change if file is renamed.

minor english correction, "Keep in mind" rather than "have in mind"

source/configuration/modules/imfile.rst
davidelang: I don't mind, just document this somewhere so that others will do the same (I generally wouldn't bother with different types of characters, but that's part of the reason I'm not the best person to write docs :-)

rgerhards: While I think it makes sense to follow the suggestion from sphinx, I know that it is just a suggestion. So I would appreciate a review what else needs to be changed, so that we can make an informed decison if this is something we really want to change.

source/configuration/modules/imfile.rst
davidelang: conceptually that's probably right, but I don't know if we want to make them all hard boundaries.

There are multiple competing goals here.

1. make it as easy as possible for people to do the common case (drives putting the required stuff first)

2. make it as easy as possible for people to find an option they have questions about (an alphabetical list of parameters is best for this)

3. avoid implying that people should tweak parameters that are there for experts

4. consistency in organization for the documentation of the different modules

5. group related parameters together (generally this breaks all the rules above :-)

6. avoid duplication of data

In this module, the multiline stuff is odd enough and involves enough different parameters with names all over the place both alphabetically and in optional/expert levels and explinations of the problems involved that apply to multiple parameters that it's probably worth having a section just for them

Expert vs Experimental is a split that I would probably only make if there are a bunch of experimental options, otherwise I'd be very tempted to lump them together (as the experimental options are almost always designed for experts)

Does this make sense?

davidelang: it may make sense to have this discussion on the mailing list.

source/configuration/modules/imfile.rst
davidelang: what is the problem with just putting the link in multiple times?

PR #294, opened 2016-12-21
WIP: Omfile.rst

mostolog:
Please, review content and some TODO's

### DIFF
--- a/source/configuration/modules/omfile.rst
+++ b/source/configuration/modules/omfile.rst
@@ -40,13 +40,6 @@ Legacy:    $ActionFileDefaultTemplate
 
 Set the default template to be used if an action is not configured to use a specific template.
 
-==============  ==================
-dirCreateMode:  "octalNumber"
-Default:        0700
-==============  ==================
-
-Sets the default directory creation mask to be used for an action if no explicit one is specified.
-
 ===============  ==================
 fileCreateMode:  "octalNumber"
 Default:         0644
@@ -81,6 +74,13 @@ Default:       current user (running Rsyslog) primary group
 =============  ============================================
 
 Sets the default file group (*gid*) to be used for an action if no explicit one is specified.
+
+==============  ==================
+dirCreateMode:  "octalNumber"
+Default:        0700
+==============  ==================
+
+Sets the default directory creation mask to be used for an action if no explicit one is specified.
 
 =========  ==============================
 dirOwner:  "username"
@@ -134,7 +134,7 @@ Default:   none
 =========  ===============
 
 For each message, the file name is generated based on the given template, having full control over how to format filenames.
-As with *file* parameter, is **strongly advised** that generated paths be like "/absolute/full/path/to/file".
+As with *file* parameter, generated paths should be like "/absolute/full/path/to/file".
 
 A cache of recent files is kept. Note that this cache can consume quite some memory (especially if large buffer sizes are used).
 For example, 1000 files with a 1M buffer and double-buffer enabled is 2GB of system memory (1000*1M*2).
@@ -211,62 +211,11 @@ On the other hand, if set to *on*, the extra headers considerably degrade compre
 Filesystem parameters
 ---------------------
 
-=========  ================
-dirOwner:  "username"
-Default:   *system default*
-Legacy:    $DirOwner
-=========  ================
-
-Set the owner for **newly created directories**. 
-This setting does not affect the owner of already existing directories.
-
-The parameter is a user name, for which the userid is
-obtained by rsyslogd during startup processing. Interim changes to
-the user mapping are not detected.
-
-================  ================
-dirOwnerNum:      "uid (number)"
-Default:          *system default*
-Available since:  7.5.8, 8.1.4
-================  ================
-
-Set the owner for **newly created directories**.
-This setting does not affect the owner of already existing directories.
-
-The parameter is a numerical ID, which is used regardless
-of whether the user actually exists. This can be useful if the user
-mapping is not available to rsyslog during startup.
-
-=========  ================
-dirGroup:  "groupname"
-Default:   *system default*
-Legacy:    $DirGroup
-=========  ================
-
-Set the group for **newly created directories**.
-This setting does not affect the group of already existing directories.
-   
-The parameter is a group name, for which the groupid is obtained by
-rsyslogd on during startup processing. Interim changes to the user
-mapping are not detected.
-
-============  ================
-dirGroupNum:  "gid (number)"
-Default:      *system default*
-============  ================
-
-Set the group for **newly created directories**.
-This setting does not affect the group of already existing directories.
-
-The parameter is a numerical ID, which is used regardless
-of whether the group actually exists. This can be useful if the group
-mapping is not available to rsyslog during startup.
-
-==========  ================
+==========  =============================
 fileOwner:  "username"
-Default:    *system default*
+Default:    *defined by module parameter*
 Legacy:     $FileOwner
-==========  ================
+==========  =============================
 
 Set the owner for **newly created files**. 
 This setting does not affect the owner of already existing files.
@@ -275,11 +224,11 @@ The parameter is a user name, for which the userid is
 obtained by rsyslogd during startup processing. Interim changes to
 the user mapping are not detected.
 
-================  ================
+================  =============================
 fileOwnerNum:     "uid (number)"
-Default:          *system default*
+Default:          *defined by module parameter*
 Available since:  7.5.8, 8.1.4
-================  ================
+================  =============================
 
 Set the owner for **newly created files**.
 This setting does not affect the owner of already existing files.
@@ -288,11 +237,11 @@ The parameter is a numerical ID, which is used regardless
 of whether the user actually exists. This can be useful if the user
 mapping is not available to rsyslog during startup.
 
-==========  ================
+==========  =============================
 fileGroup:  "groupname"
-Default:    *system default*
+Default:    *defined by module parameter*
 Legacy:     $FileGroup
-==========  ================
+==========  =============================
 
 Set the group for **newly created files**. 
 This setting does not affect the group of already existing files.
@@ -301,11 +250,11 @@ The parameter is a group name, for which the groupid is
 obtained by rsyslogd during startup processing. Interim changes to
 the user mapping are not detected.
 
-================  ================
+================  =============================
 fileGroupNum:     "gid (number)"
-Default:          *system default*
+Default:          *defined by module parameter*
 Available since:  7.5.8, 8.1.4
-================  ================
+================  =============================
 
 Set the group for **newly created files**.
 This setting does not affect the group of already existing files.
@@ -314,22 +263,14 @@ The parameter is a numerical ID, which is used regardless
 of whether the group actually exists. This can be useful if the group
 mapping is not available to rsyslog during startup.
 
-===============  ===============
+===============  =============================
 fileCreateMode:  "octalnumber"
-Default:         same as module
+Default:         *defined by module parameter*
 Legacy:          $FileCreateMode
-===============  ===============
+===============  =============================
 
 Allows to specify the file creation mask with which rsyslogd creates new files.
 If not specified, the module configuration is used.
-
-==============  ==============
-dirCreateMode:  "octalnumber"
-Default:        same as module
-Legacy:         $DirCreateMode
-==============  ==============
-
-The same as fileCreateMode, but for automatically generated directories.
 
 ===================  ===================
 failOnChOwnFailure:  "on|off"
@@ -353,6 +294,65 @@ Legacy:      $CreateDirs
 
 Create directories on an as-needed basis
 
+==============  =============================
+dirCreateMode:  "octalnumber"
+Default:        *defined by module parameter*
+Legacy:         $DirCreateMode
+==============  =============================
+
+The same as fileCreateMode, but for automatically generated directories.
+
+=========  =============================
+dirOwner:  "username"
+Default:   *defined by module parameter*
+Legacy:    $DirOwner
+=========  =============================
+
+Set the owner for **newly created directories**. 
+This setting does not affect the owner of already existing directories.
+
+The parameter is a user name, for which the userid is
+obtained by rsyslogd during startup processing. Interim changes to
+the user mapping are not detected.
+
+================  =============================
+dirOwnerNum:      "uid (number)"
+Default:          *defined by module parameter*
+Available since:  7.5.8, 8.1.4
+================  =============================
+
+Set the owner for **newly created directories**.
+This setting does not affect the owner of already existing directories.
+
+The parameter is a numerical ID, which is used regardless
+of whether the user actually exists. This can be useful if the user
+mapping is not available to rsyslog during startup.
+
+=========  =============================
+dirGroup:  "groupname"
+Default:   *defined by module parameter*
+Legacy:    $DirGroup
+=========  =============================
+
+Set the group for **newly created directories**.
+This setting does not affect the group of already existing directories.
+   
+The parameter is a group name, for which the groupid is obtained by
+rsyslogd on during startup processing. Interim changes to the user
+mapping are not detected.
+
+============  =============================
+dirGroupNum:  "gid (number)"
+Default:      *defined by module parameter*
+============  =============================
+
+Set the group for **newly created directories**.
+This setting does not affect the group of already existing directories.
+
+The parameter is a numerical ID, which is used regardless
+of whether the group actually exists. This can be useful if the group
+mapping is not available to rsyslog during startup.
+
 Expert parameters
 -----------------
 
@@ -374,7 +374,7 @@ Default:       "off"
 Legacy:        $OMFileASyncWriting
 =============  ===================
 
-if turned *on*, the files will be written in asynchronous mode via a
+If turned *on*, the files will be written in asynchronous mode via a
 separate thread. In that case, double buffers will be used so that
 one buffer can be filled while the other buffer is being written.
 Mind that if using *dynaFile* and this parameter is turned *on*, it can result in **a lot** of rsyslog threads (one per file currently open).
@@ -423,7 +423,7 @@ directory it resides after processing each batch. There currently
 is no way to sync only after each n-th batch.
 
 **Enabling sync causes a severe performance hit**.
-Actually, it slows omfile so much down, that the probability of loosing messages increases.
+Actually, it slows omfile down so much, that the probability of loosing messages increases.
 
 In short, you should enable syncing only if you know exactly what you do, and fully understand how the rest of the engine works, and have tuned the rest of the engine to lossless operations.
 
--- a/source/configuration/modules/omfile.rst
+++ b/source/configuration/modules/omfile.rst
@@ -1,434 +1,437 @@
-omfile: File Output Module
-==========================
+omfile: Text File Output Module
+###############################
 
-===========================  ===========================================================================
-**Module Name:**             **omfile**
-**Author:**                  `Rainer Gerhards <http://www.gerhards.net/rainer>`_ <rgerhards@adiscon.com>
-===========================  ===========================================================================
+================  ===========================================================================
+**Module Name:**  **omfile**
+**Author:**       `Rainer Gerhards <http://www.gerhards.net/rainer>`_ <rgerhards@adiscon.com>
+================  ===========================================================================
 
-The omfile plug-in provides the core functionality of writing messages
-to files residing inside the local file system (which may actually be
-remote if methods like NFS are used). Both files named with static names
-as well files with names based on message content are supported by this
-module.
+The omfile plugin provides the needed functionality to write messages to a local file.
 
 Configuration Parameters
-------------------------
-Omfile is a built-in module that does not need to be loaded. In order to
-specify module parameters, use
+************************
 
-::
+Omfile is a built-in module that does not need to be loaded for default settings.
+In order to specify module parameters, it must be loaded using:
 
-   module(load="builtin:omfile" ...parameters...)
-
-Note that legacy directives **do not** affect new-style RainerScript configuration
-objects. See :doc:`basic configuration structure doc <../basic_structure>` to
-learn about different configuration languages in use by rsyslog.
-
-General Notes
-^^^^^^^^^^^^^
+    module(load="builtin:omfile" [parameters])
 
 As can be seen in the parameters below, owner and groups can be set either by
-name or by direct id (uid, gid). While using a name is more convenient, using
+name or by id (*uid, gid*). While using a name is more convenient, using
 the id is more robust. There may be some situations where the OS is not able
-to do the name-to-id resolution, and these cases the owner information will be
-set to the process default. This seems to be uncommon and depends on the
+to do the name-to-id resolution and -in these cases- the owner information will be
+set to defaults. This seems to be uncommon and depends on the
 authentication provider and service start order. In general, using names
 is fine.
 
+**TODO according to @davidelang this is insecure https://github.com/rsyslog/rsyslog-doc/pull/294#pullrequestreview-14280545**
+Also, consider *octalNumber* values must always be a 4-digit octal number, with the initial digit being zero.
+Mind that actual permission depend on rsyslogd's process umask. If in doubt, use "$umask 0000" right at the beginning of the
+configuration file to remove any restrictions.
+
 Module Parameters
-^^^^^^^^^^^^^^^^^
+=================
 
-.. function::  template [templateName]
+=========  ==========================
+template:  "template-name"
+Default:   RSYSLOG_FileFormat
+Legacy:    $ActionFileDefaultTemplate
+=========  ==========================
 
-   *Default: RSYSLOG_FileFormat*
+Set the default template to be used if an action is not configured to use a specific template.
 
-   Set the default template to be used if an action is not configured
-   to use a specific template.
+==============  ==================
+dirCreateMode:  "octalNumber"
+Default:        0700
+==============  ==================
 
-.. function::  dirCreateMode [octalNumber]
+Sets the default directory creation mask to be used for an action if no explicit one is specified.
 
-   *Default: 0700*
+===============  ==================
+fileCreateMode:  "octalNumber"
+Default:         0644
+===============  ==================
 
-   Sets the default dirCreateMode to be used for an action if no
-   explicit one is specified.
+Sets the default file creation mask to be used for an action if no explicit one is specified.
 
-.. function::  fileCreateMode [default 0644] [octalNumber]
+==========  ==============================
+fileOwner:  "username"
+Default:    current user (running Rsyslog)
+==========  ==============================
 
-   *Default: 0644*
+Sets the default file owner name to be used for an action if no explicit one is specified.
 
-   Sets the default fileCreateMode to be used for an action if no
-   explicit one is specified.
+=============  ==============================
+fileOwnerNum:  "uid (number)"
+Default:       current user (running Rsyslog)
+=============  ==============================
 
-.. function:: fileOwner [userName]
+Sets the default file owner (*uid*) to be used for an action if no explicit one is specified.
 
-   *Default: process user*
+==========  ============================================
+fileGroup:  "groupname"
+Default:    current user (running Rsyslog) primary group 
+==========  ============================================
 
-   Sets the default fileOwner to be used for an action if no
-   explicit one is specified.
+Sets the default file group name to be used for an action if no explicit one is specified.
 
-.. function:: fileOwnerNum [uid]
+=============  ============================================
+fileGroupNum:  "gid (number)"
+Default:       current user (running Rsyslog) primary group 
+=============  ============================================
 
-   *Default: process user*
+Sets the default file group (*gid*) to be used for an action if no explicit one is specified.
 
-   Sets the default fileOwnerNum to be used for an action if no
-   explicit one is specified.
+=========  ==============================
+dirOwner:  "username"
+Default:   current user (running Rsyslog)
+=========  ==============================
 
-.. function:: fileGroup [groupName]
+Sets the default directory owner name to be used for an action if no explicit one is specified.
 
-   *Default: process user's primary group*
+============  ==============================
+dirOwnerNum:  "uid (number)"
+Default:      current user (running Rsyslog)
+============  ==============================
 
-   Sets the default fileGroup to be used for an action if no
-   explicit one is specified.
+Sets the default directory owner (*uid*) to be used for an action if no explicit one is specified.
 
-.. function:: fileGroupNum [gid]
+=========  ============================================
+dirGroup:  "groupname"
+Default:   current user (running Rsyslog) primary group 
+=========  ============================================
 
-   *Default: process user's primary group*
+Sets the default directroy group name to be used for an action if no explicit one is specified.
 
-   Sets the default fileGroupNum to be used for an action if no
-   explicit one is specified.
+============  ============================================
+dirGroupNum:  "gid (number)"
+Default:      current user (running Rsyslog) primary group 
+============  ============================================
 
-.. function:: dirOwner [userName]
-
-   *Default: process user*
-
-   Sets the default dirOwner to be used for an action if no
-   explicit one is specified.
-
-.. function:: dirOwnerNum [uid]
-
-   *Default: process user*
-
-   Sets the default dirOwnerNum to be used for an action if no
-   explicit one is specified.
-
-.. function:: dirGroup [groupName]
-
-   *Default: process user's primary group*
-
-   Sets the default dirGroup to be used for an action if no
-   explicit one is specified.
-
-.. function:: dirGroupNum [gid]
-
-   *Default: process user's primary group*
-
-   Sets the default dirGroupNum to be used for an action if no
-   explicit one is specified.
- 
-
+Sets the default directory group (*gid*) to be used for an action if no explicit one is specified.
+   
 Action Parameters
-^^^^^^^^^^^^^^^^^
-Note that **one** of the parameters *file* or *dynaFile* must be specified. This
-selects whether a static or dynamic file (name) shall be written to.
+=================
 
-.. function::  file [fileName]
+Note that **one and only one** of the parameters *file* or *dynaFile* must be specified.
 
-   *Default: none*
+========  ===============
+file:     "/absolute/full/path/to/file"
+Default:  none
+========  ===============
 
-   This creates a static file output, always writing into the same file.
-   If the file already exists, new data is appended to it. Existing
-   data is not truncated. If the file does not already exist, it is
-   created. Files are kept open as long as rsyslogd is active. This
-   conflicts with external log file rotation. In order to close a file
-   after rotation, send rsyslogd a HUP signal after the file has been
-   rotated away.
+This parameters set in which file will Rsyslog write. It will always write into the same file.
 
-.. function::  dynaFile [templateName]
+* If the file does not already exist, it is created.
+* If the file already exists, new data is appended to it.
 
-   *Default: none*
+**Important:** Used files remain open until rsyslog exits or receives a HUP signal.
+As this may conflict with external log file rotation, remember to send a HUP signal to Rsyslog process after the file has been rotated away.
 
-   For each message, the file name is generated based on the given
-   template. Then, this file is opened. As with the *file* property,
-   data is appended if the file already exists. If the file does not
-   exist, a new file is created. The template given in "templateName"
-   is just a regular :doc:`rsyslog template <../templates>`, so all
-   you have full control over how to format the file name.
+=========  ===============
+dynaFile:  "template-name"
+Default:   none
+=========  ===============
+
+For each message, the file name is generated based on the given template, having full control over how to format filenames.
+As with *file* parameter, is **strongly advised** that generated paths be like "/absolute/full/path/to/file".
+
+A cache of recent files is kept. Note that this cache can consume quite some memory (especially if large buffer sizes are used).
+For example, 1000 files with a 1M buffer and double-buffer enabled is 2GB of system memory (1000*1M*2).
+Files are kept open as long as they stay inside the cache and they are removed from the cache:
+
+* When a HUP signal is received.
+* Event *closeTimeout* occurs.
+* Cache runs out of space, in which case the least recently used entry is evicted.
    
-   A cache of recent files is kept. Note
-   that this cache can consume quite some memory (especially if large
-   buffer sizes are used). Files are kept open as long as they stay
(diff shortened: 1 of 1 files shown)


mostolog:
squashed, but I was asking for content review

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
-
-   *Default: 0700*
-
-   Sets the default dirCreateMode to be used for an action if no
-   explicit one is specified.
-
-.. function::  fileCreateMode [default 0644] [octalNumber]
+Also, consider *octalNumber* values must always be a 4-digit octal number, with the initial digit being zero.
+Mind that actual permission depend on rsyslogd's process umask. If in doubt, use "$umask 0000" right at the beginning of the
+configuration file to remove any restrictions.
```
davidelang: I don't like the recommendation to set umask = 0000. we should not be suggesting such insecure settings

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
-   created. Files are kept open as long as rsyslogd is active. This
-   conflicts with external log file rotation. In order to close a file
-   after rotation, send rsyslogd a HUP signal after the file has been
-   rotated away.
-
-.. function::  dynaFile [templateName]
+=================
 
-   *Default: none*
+Note that **one** of the parameters *file* or *dynaFile* must be specified.
```
davidelang: one and only one (do we generate an error if both are set? if not we should)

source/configuration/modules/omfile.rst
mostolog: I'm just a monkey typing. Ask @rgerhards for reasons. Perhaps is legacy...

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
 
-.. function::  template [templateName]
+This parameters set in which file will Rsyslog write. It will always write into the same file.
 
-   *Default: template set via "template" module parameter*
+* If the file does not already exist, it is created.
+* If the file already exists, new data is appended to it.
 
-   Sets the template to be used for this action.
+**Important:** Files are kept open as long as rsyslogd is active.
```
davidelang: The file is opened the first time that rsyslog has a log message to write out, and remains open until rsyslog exits or receives a HUP

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
 
-.. function::  closeTimeout [minutes]
+=========  ===============
+dynaFile:  "template-name"
+Default:   none
+=========  ===============
 
-   *Default: for static files: 0; for dynamic files: 10*
+For each message, the file name is generated based on the given template, having full control over how to format filenames.
+**TODO Template must be a full path?**
```
davidelang: yes, the filename and dynafile template result must both be full paths

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
-   Specifies the number of DynaFiles that will be kept open.
-   Note that this is a per-action value, so if you have
-   multiple dynafile actions, each of them have their individual caches
-   (which means the numbers sum up). Ideally, the cache size exactly
-   matches the need. You can use :doc:`impstats <impstats>` to tune
-   this value. Note that a too-low cache size can be a very considerable
-   performance bottleneck.
-
-.. function::  zipLevel [level]
+A cache of recent files is kept. Note that this cache can consume quite some memory (especially if large buffer sizes are used). Files are kept open as long as they stay inside the cache.
```
davidelang: does the cache really eat up that much memory? this sounds like it was probably created a long time ago when system memory was much smaller. The filehandle is small, what else is there? 4K per open file?

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
-   Set the file owner for files newly created. Please note that this
-   setting does not affect the owner of files already existing. The
-   parameter is a numerical ID, which which is used regardless of
-   whether the user actually exists. This can be useful if the user
-   mapping is not available to rsyslog during startup.
+People often report problems that dynafiles are not properly created.
+Apart from generating invalid paths/filenames, sometimes these problems are due to SELinux rules, which do not permit
+the creation of those files. The same happens for generic permission issues (this is often a problem
+under Ubuntu where permissions are dropped by default)
+
```
davidelang: move the dynafilecachesize parameter here, the explanation works much better in context

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
-
-   Set the file owner for files newly created. Please note that this
-   setting does not affect the owner of files already existing. The
-   parameter is a numerical ID, which which is used regardless of
-   whether the user actually exists. This can be useful if the user
-   mapping is not available to rsyslog during startup.
+People often report problems that dynafiles are not properly created.
+Apart from generating invalid paths/filenames, sometimes these problems are due to SELinux rules, which do not permit
+the creation of those files. The same happens for generic permission issues (this is often a problem
+under Ubuntu where permissions are dropped by default)
```
davidelang: replace with:

Apart from generating invalid paths/filenames, the most common cause of this is permission problems.
Permission problems can be either traditional filesystem permissions (the files need to be able to be created after any privilege drop takes place), or they can be a security module like SELinux or AppArmor.

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
+
+=========  ==========================
+template:  "template-name"
+Default:   none
```
davidelang: the default is the template specified in the module load.

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
+Legacy:    $ActionFileDefaultTemplate
```
davidelang: This actually is the equivalent of setting it at module load time. The legacy way to specify the template was
*.* /var/log/foo;templatename

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
+Specifies the number of files that will be kept open.
+Note that this is a per-action parameter, so if you have
+multiple dynafile actions, each of them have their individual caches
+(which means the numbers sum up).
+
+Ideally, the cache size exactly
+matches the need. You can use :doc:`impstats <impstats>` to tune
+this value. Note that a too-low cache size can be a very considerable
+performance bottleneck.
+
```
davidelang: it's far better to have the cache be too large than too small. Even being too small by 1 file can cripple your performance.

If the cache is too small, every write turns into:

1. find the oldest file
2. flush all data in buffers for this oldest file
3. close the oldest file
4. open the new file
5. write the data into the buffers for the new file

rgerhards:
Am 23.12.2016 10:56 schrieb "davidelang" <notifications@github.com>:

*@davidelang* commented on this pull request.
------------------------------

In source/configuration/modules/omfile.rst
<https://github.com/rsyslog/rsyslog-doc/pull/294#pullrequestreview-14281157>
:

> -
-.. function::  dynaFileCacheSize [size]
-
-   *Default: 10*
-
-   Applies only if dynamic filenames are used.
-   Specifies the number of DynaFiles that will be kept open.
-   Note that this is a per-action value, so if you have
-   multiple dynafile actions, each of them have their individual caches
-   (which means the numbers sum up). Ideally, the cache size exactly
-   matches the need. You can use :doc:`impstats <impstats>` to tune
-   this value. Note that a too-low cache size can be a very considerable
-   performance bottleneck.
-
-.. function::  zipLevel [level]
+A cache of recent files is kept. Note that this cache can consume
quite some memory (especially if large buffer sizes are used). Files
are kept open as long as they stay inside the cache.

does the cache really eat up that much memory? this sounds like it was
probably created a long time ago when system memory was much smaller. The
filehandle is small, what else is there? 4K per open file?


Depends on the settings. I have seen background on, buffer 1m, 1000 files.
That's is 2gig of system Mem (2*1m*1000).

Rainer

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/rsyslog/rsyslog-doc/pull/294#pullrequestreview-14281157>,
or mute the thread

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
+Available since:  7.3.0
+================  ========
+
+If *zipLevel* is greater than 0,
+then this setting controls if extra headers are written to make the
+resulting file extra hardened against malfunction. 
+If set to *off*, data appended to previously unclean closed files may not be
+accessible without extra tools.
+
+**TODO Note that this risk is usually expected to be bearable, and thus "off" is the default mode.**
```
davidelang: is the default off or on?

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
+=============  ===================
+
+if turned *on*, the files will be written in asynchronous mode via a
+separate thread. In that case, double buffers will be used so that
+one buffer can be filled while the other buffer is being written.
+   
+Note that in order to enable *flushInterval*, this parameter must be *on*. Otherwise, the flush interval will be ignored.
+Also note that when flushOnTXEnd is *on* but asyncWriting is *off*, output will only be written when the buffer is full.
+This may take several hours, or even require a rsyslog shutdown. However, a buffer flush can be forced in that case by sending rsyslogd a HUP signal.
+
```
davidelang: asyncwriting combined with dynafile can result is a LOT of rsyslog threads (one per file currently open)

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
+ioBufferSize:  "size"
+Default:       "4K" (KB)
+Legacy:        $OMFileIOBufferSize
+=============  ===================
+
+Size (in KB) of the buffer used to write output data. The larger the
+buffer, the potentially better performance is. The default of 4K is
+quite conservative, it is useful to go up to 64k, and 128K if you
+used gzip compression (then, even higher sizes may make sense).
+
```
davidelang: It may be worth grouping all the buffersize/flush/timeout/async settings into one section where you have a large block of text that explains how all of them interact, and then has a list of the parameters and their defaults. Right now the explanations are spread across many different parameters but interact with each other

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
-objects. See :doc:`basic configuration structure doc <../basic_structure>` to
-learn about different configuration languages in use by rsyslog.
+Omfile is a built-in module that does not need to be loaded for default settings.
+In order to specify module parameters, it must be loaded using:
 
-General Notes
-^^^^^^^^^^^^^
+    module(load="builtin:omfile" [parameters])
 
 As can be seen in the parameters below, owner and groups can be set either by
```
davidelang: this is a partial duplication of data below. It's better to either have everything in one place, or to have an explicit section for common vs expert items.

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
+obtained by rsyslogd during startup processing. Interim changes to
+the user mapping are not detected.
+
+================  ================
+dirOwnerNum:      "uid (number)"
+Default:          *system default*
+Available since:  7.5.8, 8.1.4
+================  ================
+
+**TODO why available since 2 versions? backport? bugfixes?**
```
davidelang: at that point in rsyslog development there were multiple trees being maintained, so 8.1.3 did not have this but 7.5.8 did.

yes, it was a mess and a horrible maintenance burden, that's why it's not done that way any longer :-)

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
-   **increases**. In short,
-   you should enable syncing only if you know exactly what you do, and
-   fully understand how the rest of the engine works, and have tuned
-   the rest of the engine to lossless operations.
+If different owners or groups are specified for new files or directories, and rsyslogd fails to set them:
 
-.. function::  sig.provider [providerName]
+* If set to *on*, it will log an error and NOT write to the file.
+* If it is set to *off*, the error will be ignored and processing continues.
+  Keep in mind that in the event of an error, files may be (in)accessible due to permissions.
```
davidelang: by inaccessible, this means that rsyslog may not be able to write to the file.

there should probably be a section of failure modes (disk full, permissions, etc). we talk about permissions a bit in the dynafile section, but it really applies to everything.

source/configuration/modules/omfile.rst
mostolog: If I'm not wrong it's set to 0 (false), but as documentation is inconsistent, I wanted another eyes on it.

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
-   Selects a signature provider for log signing. By selecting a provider,
-   the signature feature is turned on.
+Create directories on an as-needed basis
 
-   Currently, there are two providers available ":doc:`gt <sigprov_gt>`" and ":doc:`ksi <sigprov_ksi>`". 
+Expert parameters
+-----------------
 
-.. function::  cry.provider [providerName]
+**
```
davidelang: buffer size and flushing options are also expert-level things.

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,441 @@
...
--  people often report problems that dynafiles are not properly created.
-   The common cause for this problem is SELinux rules, which do not permit
-   the create of those files (check generated file names and pathes!). The
-   same happens for generic permission issues (this is often a problem
-   under Ubuntu where permissions are dropped by default)
+Log signing and encryption
+--------------------------
 
--  One needs to be careful with log rotation if signatures and/or
+**TODO** One needs to be careful with log rotation if signatures and/or
```
davidelang: why TODO?

rgerhards:
it's 0 (off) by default

2016-12-23 11:17 GMT+01:00 mostolog <notifications@github.com>:

source/configuration/modules/omfile.rst
```diff
@@ -444,86 +451,57 @@ Caveats/Known Bugs
...
 
-The following command writes all syslog messages into a file.
+Selects a signature provider for **log signing**. By selecting a provider, the signature feature is turned on.
 
-::
+Currently, there are two providers available ":doc:`gt <sigprov_gt>`" and ":doc:`ksi <sigprov_ksi>`". 
 
-  action(type="omfile" dirCreateMode="0700" FileCreateMode="0644"
-         File="/var/log/messages")
+**TODO `Sign log messages through signature provider Guardtime <http://www.rsyslog.com/how-to-sign-log-messages-through-signature-provider-guardtime/>`_**
```
davidelang: as noted below, there is only one provider currently implemented, but the infrastructure is designed so that others could be added easily

source/configuration/modules/omfile.rst
mostolog: I would bet yes...https://github.com/rsyslog/rsyslog/blob/1ab0d897a4f3518fc3eaf9192044eca2545c05dd/tools/omfile.c#L1231
Please, confirm!

source/configuration/modules/omfile.rst
```diff
@@ -444,86 +451,57 @@ Caveats/Known Bugs
...
+Legacy Configuration
+====================
 
 ::
 
+  # Write all mensajes into /var/log/messages
   $DirCreateMode 0700
   $FileCreateMode 0644
   *.* /var/log/messages
+
```
davidelang: shhould have an example with template format.

This is one of the modules where the legacy format gets used a lot (and if you are not needing to set anything other than file/dynafile/template is arguably clearer in the legacy format). So it may be worth putting the simple legacy example very early in the file.

source/configuration/modules/omfile.rst
mostolog: because still not done :stuck_out_tongue_closed_eyes:

davidelang:
On Fri, 23 Dec 2016, Rainer Gerhards wrote:

> +A cache of recent files is kept. Note that this cache can consume
> quite some memory (especially if large buffer sizes are used). Files
> are kept open as long as they stay inside the cache.
>
> does the cache really eat up that much memory? this sounds like it was
> probably created a long time ago when system memory was much smaller. The
> filehandle is small, what else is there? 4K per open file?
>
>
> Depends on the settings. I have seen background on, buffer 1m, 1000 files.
> That's is 2gig of system Mem (2*1m*1000).

Ok, we should specify the calculation rather than the 'scary' "it could use a 
lot of memory"

David Lang

davidelang:
Also, this isn't memory used by the cache being large, but rather by there being 
lots of open files. A cache size of 10000 doesn't hurt much if you only have 10 
files open.

rgerhards:
2016-12-23 11:27 GMT+01:00 davidelang <notifications@github.com>:

> On Fri, 23 Dec 2016, Rainer Gerhards wrote:
>
> > +A cache of recent files is kept. Note that this cache can consume
> > quite some memory (especially if large buffer sizes are used). Files
> > are kept open as long as they stay inside the cache.
> >
> > does the cache really eat up that much memory? this sounds like it was
> > probably created a long time ago when system memory was much smaller. The
> > filehandle is small, what else is there? 4K per open file?
> >
> >
> > Depends on the settings. I have seen background on, buffer 1m, 1000
> files.
> > That's is 2gig of system Mem (2*1m*1000).
>
> Ok, we should specify the calculation rather than the 'scary' "it could
> use a
> lot of memory"
>
>
There are possibly some other settings that affect memory consumption, but
these are the most important ones. Also, one thread is used per background
writer, so 1000 files --> 1000 addtl. threads.

Background writer is *really* only when we know things are very slow...
Else it is even faster to work without it.

Rainer

rgerhards:
2016-12-23 11:28 GMT+01:00 davidelang <notifications@github.com>:

> Also, this isn't memory used by the cache being large, but rather by there
> being
> lots of open files. A cache size of 10000 doesn't hurt much if you only
> have 10
> files open.
>

Right, but I remember adding this wording after seening lots of such cases.
IIRC mostly created by date in file name,so the cache *fills* up.

Rainer

davidelang:
that looks like it generates an error if neither is set, but not if both are set

davidelang:
look for setInstParamDefaults() to see what the defaults are

davidelang:
On Fri, 23 Dec 2016, mostolog wrote:

> +**TODO** One needs to be careful with log rotation if signatures and/or
>
> because still not done :stuck_out_tongue_closed_eyes:

this is not something for rsyslog to do, this is something to be implemented in 
whatever software/scripts are used for file rotation.

David Lang

source/configuration/modules/omfile.rst
mostolog: IE: Because I didn't review that part and don't understand how it works yet (IMHO mandatory for properly document it)

source/configuration/modules/omfile.rst
mostolog: Created https://github.com/rsyslog/rsyslog/issues/1346

source/configuration/modules/omfile.rst
mostolog: Ok

source/configuration/modules/omfile.rst
mostolog: Ok.

source/configuration/modules/omfile.rst
mostolog: Ok.

rgerhards:
2016-12-23 12:30 GMT+01:00 mostolog <notifications@github.com>:

> *@mostolog* commented on this pull request.
> ------------------------------
>
> In source/configuration/modules/omfile.rst
> <https://github.com/rsyslog/rsyslog-doc/pull/294>:
>
> >
> -   *Default: for static files: 0; for dynamic files: 10*
> +For each message, the file name is generated based on the given template, having full control over how to format filenames.
> +**TODO Template must be a full path?**
>
> Ok.
>
>

To be precise, this is not necessary from a technical point of view.
rsyslog uses relative path names, and applies them from whereever it's home
directory is set to (usually the file system root). So in uncommon
installs, a relative path may make sense. But in 99.999999% of the cases,
an absolute path is the best (and safe) bet.

Rainer

source/configuration/modules/omfile.rst
mostolog: Ok.

source/configuration/modules/omfile.rst
mostolog: Not obvious $ActionFileDefaultTemplate was for module...Ok

source/configuration/modules/omfile.rst
mostolog: Ok.

source/configuration/modules/omfile.rst
mostolog: Ok.

source/configuration/modules/omfile.rst
mostolog: Ok.

source/configuration/modules/omfile.rst
mostolog: Ok. I'll group them now and I'll try to get the bigger picture.

source/configuration/modules/omfile.rst
mostolog: It will be good, for other doc pages, to know what's the current release/supported process.

source/configuration/modules/omfile.rst
mostolog: Ok.

source/configuration/modules/omfile.rst
mostolog: What's background on...ie: where the 2 came from?

**2**\*1m\*1000

source/configuration/modules/omfile.rst
mostolog: merged

source/configuration/modules/omfile.rst
mostolog: Ok.

source/configuration/modules/omfile.rst
mostolog: Agree. I'm just reviewing not remaking docs while I'm learning to use each features. Probably that could be a future improvement.

Anyway, I'm open for suggestions on how to do it ;)

source/configuration/modules/omfile.rst
mostolog: Ok.

mostolog:
Updated again. Pending:
- _$umask 0000 is insecure_ **comment**
- _$F$OMFileForceCHOwn equivalent to the "ForceChOwn" parameter._ **ForceChOwn is undocumented, and I don't know if that $F is a typo**
- _$ActionFileEnableSync equivalent to the "enableSync" parameter._ **enableSync is undocumented.**

I still have to review signing/encryption...

Is there any way (apart from https://github.com/rsyslog/rsyslog-doc#instructions) to get a sphinx/web preview? A dummy test page integrated into CI building that we can easily change, gets build by sphinx and published under rsyslog.com for preview will be more than welcome...

rgerhards:
2016-12-23 13:11 GMT+01:00 mostolog <notifications@github.com>:

> *@mostolog* commented on this pull request.
> ------------------------------
>
> In source/configuration/modules/omfile.rst
> <https://github.com/rsyslog/rsyslog-doc/pull/294>:
>
> > -
> -.. function::  dynaFileCacheSize [size]
> -
> -   *Default: 10*
> -
> -   Applies only if dynamic filenames are used.
> -   Specifies the number of DynaFiles that will be kept open.
> -   Note that this is a per-action value, so if you have
> -   multiple dynafile actions, each of them have their individual caches
> -   (which means the numbers sum up). Ideally, the cache size exactly
> -   matches the need. You can use :doc:`impstats <impstats>` to tune
> -   this value. Note that a too-low cache size can be a very considerable
> -   performance bottleneck.
> -
> -.. function::  zipLevel [level]
> +A cache of recent files is kept. Note that this cache can consume quite some memory (especially if large buffer sizes are used). Files are kept open as long as they stay inside the cache.
>
> What's background on...ie: where the 2 came from?
>
> *2**1m*1000
>
>
>
Background wiriting uses double buffers. Not sure if there is even a
parameter to specify the number of buffers (so n >= 2),

Rainer

rgerhards:
2016-12-23 13:44 GMT+01:00 mostolog <notifications@github.com>:

> Updated again. Pending:
>
>    - *$umask 0000 is insecure* *comment*
>    - *$F$OMFileForceCHOwn equivalent to the "ForceChOwn" parameter.* *ForceChOwn
>    is undocumented, and I don't know if that $F is a typo*
>
>
$F is wrong

>
>    - *$ActionFileEnableSync equivalent to the "enableSync" parameter.* *enableSync
>    is undocumented.*
>
> I still have to review signing/encryption...
>
> Is there any way (apart from https://github.com/rsyslog/
> rsyslog-doc#instructions) to get a sphinx/web preview? A dummy test page
> integrated into CI building that we can easily change, gets build by sphinx
> and published under rsyslog.com for preview will be more than welcome..
>

I admit I do not understand why you object so much against generating the
doc on your machine. This step is mandatory in any case, and you can see
this if you look at the CI results of your contributions. They are all
broken. While I know you deeply involved in the contribution, it makes you
look very unprofessional. The right workflow is to make changes, check that
they work, and then do a pull request. This also applies to source code.
PRs should **not** be done on any untested element.

Rainer

mostolog:
>I admit I do not understand why you object so much against generating the doc on your machine. This step is mandatory in any case, and you can see this if you look at the CI results of your contributions. They are all broken. While I know you deeply involved in the contribution, it makes you look very unprofessional. The right workflow is to make changes, check that they work, and then do a pull request. This also applies to source code. PRs should **not** be done on any untested element.

Ok. I added the WIP prefix to reflect that this wasn't finished and requested content review. Same for other PR.

Apart from those comments,

- What do you think about parameter type/default/legacy format?
- Do you think this reviewing process worths the effort? (IMHO: it would be much noticeable in other pages like rulesets, queues, actions...)

I'll appreciate any feedback.

rgerhards:
before commenting further, I would really like to know why you insist on not testing your changes...

davidelang:
On Fri, 23 Dec 2016, mostolog wrote:

> It will be good, for other doc pages, to know what's the current release/supported process.

There should be a separate page on this (so that it's not in multiple places 
when it needs to be changed)

the current policy is that we support the latest release only.

The best place for support is the mailing list, the web forums are monitored, 
but not by as many people.

David Lang

mostolog:
any more feedback?

source/configuration/modules/omfile.rst
davidelang: I would just leave out these comments here and put them in the user/group sections

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,437 @@
...
+Default:   RSYSLOG_FileFormat
+Legacy:    $ActionFileDefaultTemplate
+=========  ==========================
 
-   *Default: process user*
+Set the default template to be used if an action is not configured to use a specific template.
 
-   Sets the default fileOwnerNum to be used for an action if no
-   explicit one is specified.
+==============  ==================
```
davidelang: move this so that it's with all the other dir options

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,437 @@
...
-   explicit one is specified.
-
-.. function:: fileOwner [userName]
-
-   *Default: process user*
-
-   Sets the default fileOwner to be used for an action if no
-   explicit one is specified.
+Module Parameters
+=================
```
davidelang: I would add a note that these seldom need to be changed, and would consider moving the entire module parameters section below the action section (on the basis that most people won't need to see it), but as a consistency thing, it may make sense to keep it here (possibly with a link to skip down to the action parameters)

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,437 @@
...
-   A cache of recent files is kept. Note
-   that this cache can consume quite some memory (especially if large
-   buffer sizes are used). Files are kept open as long as they stay
-   inside the cache.
-   Files are removed from the cache when a HUP signal is sent, the
-   *closeTimeout* occurs, or the cache runs out of space, in which case
-   the least recently used entry is evicted.
-
-.. function::  template [templateName]
+**Important:** Used files remain open until rsyslog exits or receives a HUP signal.
```
davidelang: not just remain open, rsyslog will keep writing to the file, even if it's renamed.

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,437 @@
...
 
-   *Default: template set via "template" module parameter*
+=========  ===============
+dynaFile:  "template-name"
+Default:   none
+=========  ===============
 
-   Sets the template to be used for this action.
+For each message, the file name is generated based on the given template, having full control over how to format filenames.
+As with *file* parameter, is **strongly advised** that generated paths be like "/absolute/full/path/to/file".
```
davidelang: just say that they should be absolute paths.

source/configuration/modules/omfile.rst
davidelang: I would consider something along the lines of a table listing the option names and then a  link pointing at the same parameters that can be set in the action.

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,437 @@
...
+accessible without extra tools, something usually bearable.
+On the other hand, if set to *on*, the extra headers considerably degrade compression, resulting in files four to five times bigger.
+
+Filesystem parameters
+---------------------
+
+=========  ================
+dirOwner:  "username"
+Default:   *system default*
+Legacy:    $DirOwner
```
davidelang: consider adding a note to this that says that the default can be set in the module load

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,437 @@
...
+
+This option modifies behaviour of file creation.
+
+If different owners or groups are specified for new files or directories, and rsyslogd fails to set them:
+
+* If set to *on*, it will log an error and NOT write to the file.
+* If it is set to *off*, the error will be ignored and processing continues.
+  Keep in mind that in the event of an error, rsyslog may not be able to write on file.
+
+===========  ===========
```
davidelang: dir parameters should be listed after this option

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,437 @@
...
--  One needs to be careful with log rotation if signatures and/or
+if turned *on*, the files will be written in asynchronous mode via a
+separate thread. In that case, double buffers will be used so that
+one buffer can be filled while the other buffer is being written.
+Mind that if using *dynaFile* and this parameter is turned *on*, it can result in **a lot** of rsyslog threads (one per file currently open).
+   
+Note that in order to enable *flushInterval*, this parameter must be *on*. Otherwise, the flush interval will be ignored.
+Also note that when flushOnTXEnd is *on* but asyncWriting is *off*, output will only be written when the buffer is full.
+This may take several hours, or even require a rsyslog shutdown. However, a buffer flush can be forced in that case by sending rsyslogd a HUP signal.
+
```
davidelang: possibly add a note that the most common reason to use this is to improve performance of zipped files

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,437 @@
...
+========  ========
+
+Enables file syncing capability.
+
+When enabled, rsyslog does a sync to the data file as well as the
+directory it resides after processing each batch. There currently
+is no way to sync only after each n-th batch.
+
+**Enabling sync causes a severe performance hit**.
+Actually, it slows omfile so much down, that the probability of loosing messages increases.
```
davidelang: change to "Actually, it slows omfile down so much"

source/configuration/modules/omfile.rst
```diff
@@ -1,434 +1,437 @@
...
+
+In short, you should enable syncing only if you know exactly what you do, and fully understand how the rest of the engine works, and have tuned the rest of the engine to lossless operations.
+
+Log signing and encryption
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**TODO I still have to review this**
```
davidelang: need to eliminate TODO items before final submission

source/configuration/modules/omfile.rst
mostolog: doh!

source/configuration/modules/omfile.rst
mostolog: For better documenting/understanding:
Files are used through inodes for writing, hence renaming the file won't affect Rsyslog.
I guess rotation is renaming+creating a new with the older-name. is that correct?

Does rsyslog restart/HUP "reset" the currently inode list or something like that?

Anyway, if rotation happens with rsyslog running, some messages could go to rotated file, instead of new one, isnt it?

To sum up: I'm starting to think I didn't understood this properly. could you...?

source/configuration/modules/omfile.rst
mostolog: Ok

source/configuration/modules/omfile.rst
mostolog: changed to "defined by module parameter" could also use "inherited from..."

source/configuration/modules/omfile.rst
mostolog: Ok

source/configuration/modules/omfile.rst
mostolog: please, be more specific, as I didn't understood what exactly improve zipped files performance

source/configuration/modules/omfile.rst
mostolog: thx!

mostolog:
> @davidelang I would just leave out these comments here and put them in the user/group sections

you mean in both sections? I'll try to do it using rst+sphinx "like a reference". otherwise, having paragraphs "twice" it's not a good idea when documenting

davidelang:
On Thu, 12 Jan 2017, mostolog wrote:

> For better documenting/understanding:
> Files are used through inodes for writing, hence renaming the file won't affect Rsyslog.
> I guess rotation is renaming+creating a new with the older-name. is that correct?
>
> Does rsyslog restart/HUP "reset" the currently inode list or something like that?
>
> Anyway, if rotation happens with rsyslog running, some messages could go to rotated file, instead of new one, isnt it?
>
> To sum up: I'm starting to think I didn't understood this properly. could you...?

ok, going into the gory details:


On any *nix system, filenames are used to open a file (and multiple filenames 
can point at the same file, softlinks are a pointer to another file, hardlinks 
are just another name for the same file, with no "real" filename, all names are 
equally valid".

Names point at inodes for the file (slight simplification)

Once a file is opened and being written to, it's accessed by the program via a 
filehandle, which access the file data via the inode.

When you rotate a log file, it can be done multiple ways, the "proper" method 
[1] is to rename the file to a new name, and then create a new file. When you 
renmae the file, you are technically creating a new hardlink to the file and 
then removing the old hardlink (the OS guarantees that with the mv command, you 
will always have one, and only one of the two names valid at any point in time, 
unless you crash the system in the middle)

*nix filesystems won't actually fully delete a file that a program has open. You 
can delete all the names pointing at the file, but programs that have the file 
open can still see all the contents and the disk space will not be freed. Some 
programs create temporary files by opening and then deleting them. When the 
program exits, the OS will know that there is nothing that has the file open and 
will fully delete the file at that time.

So after a file is renamed, anything that has the file open will continue to 
access it without knowing about the rename.

This is true for reading the file (i.e. imfile) as well as writing to the file 
(omfile). To detect that the file has been renamed, the program needs to 
remember what inode it found when it looked up the filename in the first place, 
then periodically lookup the filename again, if the inode is different from what 
the program remembers, the program knows that something changed.

for omfile, rsyslog doesn't bother to do this. Instead it just keeps writing to 
the file it has open until it receives a HUP signal. At that point it closes all 
outputs (files and network connections), and then re-opens them as it needs to 
for new log messages that arrive.

unfortunantly, signal processing is asynchronous, so to properly rotate a log 
file, you have to:

1. rename the current log file (new log messages that arrive keep going to the 
old file)

2. (for some programs, not rsyslog) create the new log file

3. send a HUP to the process that is writing to the file

4. wait 'long enough' (a few seconds is usually enough) for the program to 
notice the HUP and close the file that you renamed. The program will now start 
writing to the new file.

5. now you can process the log file

what rsyslog does in imfile is to periodically check if the filename now points 
to a different inode. If it does, it finishes reading the file it's on (if it's 
in the middle of a multi-line log it considers it completed), closes the file.

David Lang

[1] what some programs do instead of the proper file rotation is that they 
delete the contents of the file and start writing again from the beginning. They 
assume that anything that wanted the old contents copied it first, but there is 
a race condition between one programe writing to the file and then truncating it 
and another file trying to make a copy of it. It doesn't work well, but that is 
the reason for implementing the experimental truncation detection functionality 
in imfile, to deal with programs that do this wrong

davidelang:
On Thu, 12 Jan 2017, mostolog wrote:

>> --  people often report problems that dynafiles are not properly created.
> -   The common cause for this problem is SELinux rules, which do not permit
> -   the create of those files (check generated file names and pathes!). The
> -   same happens for generic permission issues (this is often a problem
> -   under Ubuntu where permissions are dropped by default)
> -
> --  One needs to be careful with log rotation if signatures and/or
> +if turned *on*, the files will be written in asynchronous mode via a
> +separate thread. In that case, double buffers will be used so that
> +one buffer can be filled while the other buffer is being written.
> +Mind that if using *dynaFile* and this parameter is turned *on*, it can result in **a lot** of rsyslog threads (one per file currently open).
> +
> +Note that in order to enable *flushInterval*, this parameter must be *on*. Otherwise, the flush interval will be ignored.
> +Also note that when flushOnTXEnd is *on* but asyncWriting is *off*, output will only be written when the buffer is full.
> +This may take several hours, or even require a rsyslog shutdown. However, a buffer flush can be forced in that case by sending rsyslogd a HUP signal.
> +
>
> please, be more specific, as I didn't understood what exactly improve zipped files performance

compression works by detecting redundant data in the output and writing out a 
shortcut version of it (think of it as saying "the data here is the 30 
characters that occured 500 characters ago).

Up to the limits of the compression algorithm, the bigger the chunk of data that 
the compression algorithm can see at once, the better it's compression is going 
to be as it can use more pointers and less data.

So if your buffer is only 4K, and your messages are 512 bytes, you can only look 
at the last 8 messages when trying to compress the data. you will see some wins 
(dates and hostnames are going to be common, even if the log data isn't)

but if your buffer is 512K, you can look at the last 1k messages and find much 
more data that can be replaced by pointers.

David Lang

davidelang:
On Thu, 12 Jan 2017, mostolog wrote:

> you mean in both sections? I'll try to do it using rst+sphinx "like a 
> reference". otherwise, having paragraphs "twice" it's not a good idea when 
> documenting

I agree, but the entire permission section is basically duplicated already. 
Which is why I suggested that in the module section, you just list the names of 
the parameters and have a link to the action permission section so that all the 
data and explinations are only listed once.

David Lang

PR #297, opened 2017-01-13
Update description of template SyslogProtocol23Format

dud225:
Hello

as far as I understand, RFC5424 is the actual syslog standard and is implemented by template _SyslogProtocol23Format_.

Let me know if you agree with thiis template description.


rgerhards:
Thanks for taking the time to look at this.

Some fine nits: this format actually implement the 23 draft, not the final standard. But both are very very close. We couldn't change as things were in production for quite some time when RFC5424 was finally approved.

Also, the not on other syslogd's is quite pessimistic - as far as I have seen and heared, most current syslogd's nowadays understand that format.

dud225:
Hello Rainer

thanks for your feedback, I've updated my PR accordingly.

rgerhards:
Thx!

PR #330, opened 2017-05-02
Added the description of the Guardtime KSI module rsyslog-ksi-ls12

kadriiiii:
Created short description for the new Guardtime KSI module rsyslog-ksi-ls12 that will be available since rsyslog 8.27.


kadriiiii:
Removed the source/_ext/edit_on_github.pyc

rgerhards:
would it be possible to squash the commits? If not, I can merge as-is, it's not such a problem for the doc project. But a squash would look nicer.

PR #348, opened 2017-06-26
Issue #1629: implement 'fileoffset' as imfile metadata

derekjsonar:


rgerhards:
better late than never ... sorry for screwing up ;-)

derekjsonar:
No screw up at all! I went ahead and deleted the branch.

PR #369, opened 2017-10-06
Various typo and wording tweaks

deoren:
I'm new to making contributions via GitHub, so please let me know if I'm going about this correctly.

Thanks.

### DIFF
--- a/source/rainerscript/global.rst
+++ b/source/rainerscript/global.rst
@@ -171,6 +171,20 @@ The following parameters can be set:
   syslogtag. In those cases, the programname is truncated at the
   first slash. If this setting is changed to "on", slashes are
   permitted and will not terminate programname parsing.
+  
+  - **parser.permitSlashInProgramName** [on/off] available in 8.25.0+
+
+  **Default:** off
+
+  This controls whether slashes in the static part of the tag are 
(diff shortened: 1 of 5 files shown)


deoren:
I actually meant to include the changes to the global object documentation as a separate pull request, but apparently I forgot to branch and make those changes in a separate branch. Now it is showing as an addition to the first PR instead of a separate item.

If I need to I can blow away my fork and redo the work.

rgerhards:
Thanks for the PR. I think multiple commits are fine, actually it's along the same theme (multiple PRs would also be fine for me, just for the records ;-)).

deoren:
Thanks @rgerhards . Going forward I'll make sure to create separate branches for changes.

source/rainerscript/global.rst
```diff
@@ -171,6 +171,20 @@ The following parameters can be set:
   syslogtag. In those cases, the programname is truncated at the
   first slash. If this setting is changed to "on", slashes are
   permitted and will not terminate programname parsing.
+  
+  - **parser.permitSlashInProgramName** [on/off] available in 8.25.0+
```
deoren: Note to self: Submit a new PR that fixes the spacing here.

PR #372, opened 2017-10-07
Added missing word to the explanation of the set statement

deoren:


rgerhards:
The patches look fine, except for the first commit "Merge...". This makes the commit history really clumpsy and should usually not happen.

In general, you should have your own master branch, that you only update with new work from the official one. That way, all merges should be fast-forward. Out of that branch, you should branch off the feature branches.

Would you be willing to redo the commits by creating new feature branches and cherry-picking the actual commit to it? If not, I can cherry-pick, but that will make all PRs look like they were not accepted.

deoren:
Sure, I can do that. Thanks for your patience as I try to learn what I'm doing with GitHub. :)

deoren:
> @rgerhards In general, you should have your own master branch, that you only update with new work from the official one. That way, all merges should be fast-forward. Out of that branch, you should branch off the feature branches.

Do you happen to know whether all of those steps are possible using just the web UI? Just curious. I hope to give this a shot later today (regardless of the approach needed).

rgerhards:
> Do you happen to know whether all of those steps are possible using just the web UI? Just curious. I hope to give this a shot later today (regardless of the approach needed).

I think there is an option in the github UI to not log fast forward merges. With that, you should probably be able to do all via the UI. But I am not 100% sure, I use the command line.

deoren:
Thanks for the info. Looking over the Git documentation it sounds like this is referred to as "rebasing". Is that correct?

Thanks for your help.

PR #410, opened 2017-12-07
Restructure and enhancements

rgerhards:
this includes PR https://github.com/rsyslog/rsyslog-doc/pulls
and has some enhancements to the description.


rgerhards:
@deoren I wouldn't mind if you have a look ;-) It's far from final, but I think we can merge, as it doesn't make matters worse (IMHO it even is a slight improvement). My plan right now is to do a bit "here and there" as I find time. I wouldn't like to wait until 100% completion but rather add each step when it is useful. Even if that means git version history clutter...

deoren:
> @rgerhards: I wouldn't mind if you have a look ;-) 

Generating an epub version of your branch in order to review the changes. I'll try to give some feedback in the next day or so.

> It's far from final, but I think we can merge, as it doesn't make matters worse (IMHO it even is a slight improvement). My plan right now is to do a bit "here and there" as I find time. I wouldn't like to wait until 100% completion but rather add each step when it is useful. Even if that means git version history clutter...

That is my situation as well. With the time I have to look at this, it's easier to make quick small changes and push back vs trying to save them up for a larger, more cohesive PR later.

rgerhards:
note that the sysklogd description is actually an import from the (sysklogd) man page -- there is much more to do.

deoren:
@rgerhards I can already tell that I really like the new index. It really opens up the specific topics for discovery.

deoren:
@rgerhards I've been reviewing the changes and all in all, I like them. I think the new layout provides a good base to build on.

I see some duplication between some of the content:

- https://github.com/rgerhards/rsyslog-doc/blob/restructure/source/configuration/conf_formats.rst
- https://github.com/rgerhards/rsyslog-doc/blob/restructure/source/configuration/basic_structure.rst

but, I'm guessing this is intentional (for the moment) to test how the content flows.

rgerhards:
> but, I'm guessing this is intentional (for the moment) to test how the content flows.

Yes. TBH, I think the current state of the doc is so bad that it almost is impossible to make it much worse ;-). So I opted to apply the changes, but leave existing stuff where it currently is. As we go forward, we can remove it from those other places.

My initial target is to get the structure definition somewhat clearer. That probably means a bit more of duplication followed then by cleanup. I am actually undecided but think we can use quite a bit of the old text as basis, but need to streamline it and shuffle it around.

Issue #391, opened 2017-11-17
rsyslog documentation doesn't explicitly state that configuration parameters are NOT case sensitive


deoren:
Unless I overlooked it somewhere, I didn't see it explicitly mentioned that the configuration parameters within the ryslog configuration file are not case sensitive.

I plan to review the documentation more thoroughly and submit a PR so that the docs clearly state that this is the case. If it's already stated elsewhere, I will replicate that statement to other portions of the documentation to help emphasize the fact.

http://www.rsyslog.com/doc/v8-stable/configuration/basic_structure.html

**EDIT**

Confirmation re case-insensitive nature of configuration parameters given here:

http://lists.adiscon.net/pipermail/rsyslog/2017-November/044971.html

deoren:
@rgerhards Just in case I missed it, are legacy configuration statements case-sensitive?

rgerhards:
no

deoren:
@rgerhards  Thanks for confirming.

deoren:
@rgerhards Are the phrases "legacy configuration statements" and "legacy configuration directives" interchangeable?

rgerhards:
yes. Now that you said this, I think we should really try to use these terms (config format name) consistently. I think we should base it on this here: http://www.rsyslog.com/doc/master/configuration/conf_formats.html

Just be aware that I will probably slightly rename the "sysklogd" format as it may make sense to include some of the very basic enhancements done in rsyslog. That tripple has the advantage that it also is very easy to explain what is good to use(sysklogd/RainerScript) and what is really bad "legay"). Again, I'll probably propose some better names for the tripple.

deoren:
> @rgerhards: Now that you said this, I think we should really try to use these terms (config format name) consistently. 

Thanks for the details.

Just to make sure I'm understand, are you saying that ...

- (1) we should use the same config format name for both legacy and RainerScript formats
- (2a) that we should consistently use the config format name of `legacy rsyslog` for the legacy format throughout the documentation
- (2b) we refer to the new config format name as RainerScript throughout the documentation

I assume that you mean the second option?

On a related note, I notice that there are different names for the parameters:

- http://www.rsyslog.com/doc/v8-stable/configuration/modules/omfile.html (`configuration parameters`)
- http://www.rsyslog.com/doc/v8-stable/configuration/modules/ompipe.html (`configuration directives`)

Is that desired?

[...]

PR #414, opened 2017-12-10
Prototype substitutions: terms, doc release ver

deoren:
Submitting this PR for feedback in case having a PR makes the changes easier to discuss than referencing a commit id on the related issue (rsyslog/rsyslog-doc#391).

@davidelang @rgerhards


rgerhards:
I like the new version. We just need to check that it is easy to generate a regular version ("8.32.0"). Haven't tried that. Replacements are also great. But I am not sure if it is so useful to do them for the config format names. First of all, I'd like to have hyperlink capabiltity but more important I think the rest of the text needs to match the name used. I'll submit an experimental PR so show what I mean.

rgerhards:
Please have a look at https://github.com/rsyslog/rsyslog-doc/pull/417

deoren:
> @rgerhards: but more important I think the rest of the text needs to match the name used

I agree. 

The formats page was just the prototype. The next step would be to update all other pages to use the substitution references for later expansion.

deoren:
> @rgerhards: Please have a look at #417

Will do.

rgerhards:
> The next step would be to update all other pages to use the substitution references for later expansion.

I didn't make myself clear. I don't mean the other pages. I mean the text the comes together with the format name. E.g. if we would call it "simple format" than the text would be along the lines of 

"You use simple format to do simple things in rsyslog configuration."

with "basic" format, it would be

"You use basic format to perform basic configuration tasks (but not advanced ones)."

So I think simple vs. basic needs to become the full "theme" of the text. You cannot do this with substitutions, as the overall text is slightly different. I have done this in #417, so watch out for the fine details (if I managed to get that done as a non-native speaker ;-)).

Does that clarify?

deoren:
> @rgerhards: I didn't make myself clear. I don't mean the other pages. I mean the text the comes together with the format name. E.g. if we would call it "simple format" than the text would be along the lines of
> 
> "You use simple format to do simple things in rsyslog configuration."
> 
> with "basic" format, it would be
> 
> "You use basic format to perform basic configuration tasks (but not advanced ones)."
> 
> So I think simple vs. basic needs to become the full "theme" of the text. You cannot do this with substitutions, as the overall text is slightly different. 

I think I follow. Is your intention of a "theme" for all pages or just this one? It may be a natural fit for the "triple" as you called it, but I'm not sure that pattern will work as well elsewhere. Provided that you are not looking to update all pages to use that theme, then substitutions could work well in other places.

For example, at present many of the module pages have a "Legacy Configuration Parameters" section at the bottom. I feel that something like this:

```
|FormatsObsoleteNewName| Configuration Directives
-------------------------------------------------

**Legacy directives should NOT be used when writing new configuration files.**

Note that the |FormatsObsoleteNewName| configuration parameters do **not** affect
new-style definitions via the input() object. This is
by design. To set defaults for input() objects, use module parameters
in the
```

could work. It is some work to go through and make the replacements, but this particular example doesn't have a theme. I think it may make sense to have such a theme in the `source/configuration/conf_formats.rst` file, but maybe not the other pages.

> @rgerhards: I have done this in #417, so watch out for the fine details (if I managed to get that done as a non-native speaker ;-)).

I think you do a great job, especially as a non-native speaker. I still struggle with English and it's my first (and only) language.

That said, I think any non-native speaker will usually benefit from having a native speaker proof their work for minor adjustments. Theoretically open source software helps in that regard. :)

rgerhards:
> I think I follow. Is your intention of a "theme" for all pages or just this one? It may be a natural fit for the "triple" as you called it, but I'm not sure that pattern will work as well elsewhere. Provided that you are not looking to update all pages to use that theme, then substitutions could work well in other places.

I think it's for "all" pages ... of course not really "all" but all where there is a description. I agree that susbstitutions make great sense for many things, I am just skeptic about the names. Let's take a hypothetical sample of a page describing omfile. With "simple" it would probably be along these lines:
```
Most often, you simply want to write a file, without any enhanced attributes. In that case, simple configuration format is fully sufficient. For more enhanced things, like setting permissions, enhanced format (described below) shall be used.
```
and with "basic"
```
Most often, you basically want to write a file, without any advanced attributes. In that case, basic configuration format is fully sufficient. For more advanced things, like setting permissions, advanced format (described below) shall be used.
```
"basically" sound and probably is wrong here, but I guess you get the idea. I think there are many areas where we need to describe the simple/enhanced vs. basic/advanced theme... But for sure not everywhere.

deoren:
> @rgerhards: I like the new version. We just need to check that it is easy to generate a regular version ("8.32.0"). Haven't tried that. 

Here is the bash script I was (am?) using prior to the `conf.py` adjustments to automatically include the info:

```
# Pull details dynamically
version=$(grep -E '^version' source/conf.py | awk '{print $NF}' | sed "s/'//g")
datestamp=$(date '+%Y%m%d')
commit=$(git describe | awk -F'-' '{print $NF}')
release="${version}.${datestamp}.${commit}"

declare -a formats
formats=(
  epub
  html
)

for format in "${formats[@]}"
do
    sphinx-build -D version="$version" -D release="$release" -b $format source build
done
```

If we instead go with the prototype `conf.py` file, you end up with the `rsyslog 8.32-20171210-a8eb054 documentation` string in various places in the documentation (across the top in HTML format, across the cover in epub). If your goal was to display on the version information for "stable" builds (i.e., drop the date and commit hash), you could call `sphinx-build` like this:

`sphinx-build -D release="8.32.0" -b $format source build`

Unless there is a way to have `sphinx-build automatically refer to an existing value in the `conf.py` file,  you're having to specify it again with the `sphinx-build` command. I'm new to Sphinx, so I'm probably missing something.

Perhaps we should have the build script do that lifting instead of embedding the logic in the `conf.py` file?

deoren:
Stepping back, I think we share several goals with what we have prototyped thus far:

- Update all documentation to consistently use the same names for the three supported formats

- As much as we are able, clearly state that the legacy format should only be used as a bridge to get from an older configuration to a newer one that uses either of the two preferred (and fully supported) formats

- Mention that the older, multi-decade old format (aka, "syslogd") is still fully supported and is recommended for simpler tasks where one line is all that is needed to accomplish a task

- Mention the newer syntax (I keep trying to say "current", but if the older "syslogd" syntax is still recommended then it is "current" also) and emphasize that it is best for anything more than a one-liner.

> @rgerhards: I think it's for "all" pages ... of course not really "all" but all where there is a description. I agree that susbstitutions make great sense for many things, I am just skeptic about the names. Let's take a hypothetical sample of a page describing omfile. With "simple" it would probably be along these lines:
> 
> > Most often, you simply want to write a file, without any enhanced attributes. In that case, simple configuration format is fully sufficient. For more enhanced things, like setting permissions, enhanced format (described below) shall be used.
> and with "basic"
> 
> > Most often, you basically want to write a file, without any advanced attributes. In that case, basic configuration format is fully sufficient. For more advanced things, like setting permissions, advanced format (described below) shall be used.
>
> "basically" sound and probably is wrong here, but I guess you get the idea. I think there are many areas where we need to describe the simple/enhanced vs. basic/advanced theme... But for sure not everywhere.

I think we can use a substitution reference for the format names, but still have the content read well and add emphasis where needed to teach best practices for using (or not in the case of "legacy") the available formats.

To use your examples (with slightly altered wording):

> More often than not, you just want to write log messages to a file without any additional attributes. For those situations, the `basic` configuration format is fully sufficient. For more complex things, like setting permissions, the `advanced` configuration format (described below) should be used.

> More often than not, you just want to write log messages to a file without any additional attributes. For those situations, the `simple` configuration format is fully sufficient. For more complex things, like setting permissions, the `advanced` configuration format (described below) should be used.

With a substitution reference, we would have this:

> More often than not, you just want to write log messages to a file without any additional attributes. For those situations, the |FormatsBasicNewName| configuration format is fully sufficient. For more complex things, like setting permissions, the |FormatsAdvancedNewName| configuration format (described below) should be used. As noted previously, you should not use the |FormatsObsoleteNewName| format in new configurations.

I've not tested how to use hyperlinking with substitutions. I understand that may be problematic if support for doing so is not available. I'll look into that and report back when able.

deoren:
I meant to post this earlier:

```
deoren@Hope MINGW64 ~/Documents/dev/github/rsyslog-doc/source (master)
$ grep -ri "legacy" . | wc -l
162
```

```
deoren@Hope MINGW64 ~/Documents/dev/github/rsyslog-doc/source (master)
$ grep -ri "rainerscript" . | wc -l
81
```

```
deoren@Hope MINGW64 ~/Documents/dev/github/rsyslog-doc/source (i-391-conf-formats)
$ grep -ri "sysklogd" . | wc -l
46
```


I've not dug into the content in detail to confirm that all of those words are used specifically to describe the three formats, but I suspect that the format name are used much more than initially thought.

rgerhards:
I think let's go ahead with the substitutions and see where this leads us to. :-)  can you prepare the PR so that I can merge it (or just say it is ready)?

deoren:
Alright. Did you want me to rebase off of your recent branch or just go with what we have here?

I can also break out the changes to conf.py separately if you want.

rgerhards:
Let's go with what we have here. I can always create a new PR for my stuff
if that makes sense. If I get into trouble with the version, I may need
your help.

Sent from phone, thus brief.

deoren:
Alright. Sounds good then: merge this and then refine with follow-up PRs.

rgerhards:
I think I'll shorten the substitution names a bit after the merge, just so that we can safe typing...

deoren:
> I think I'll shorten the substitution names a bit after the merge, just so that we can safe typing...

Alright. This were mostly that verbose for illustration purposes, but I do tend to lean that direction at times.

PR #416, opened 2017-12-10
Update terms, emphasize case-insensitive parameters

deoren:
My editor "helped" by trimming whitespace, so some NOOP changes were rolled into this commit in addition to the intended changes.

It feels like overkill, but I ended up placing this statement in the majority of files I found that mentioned parameters/directives:

`Note: parameter names are case-insensitive.`

If that ends up being useful, perhaps we could use a substitution reference for it as well.

While I was making that change I tried to replace instances of "directive" with "parameter" as discussed on the linked issue.

refs rsyslog/rsyslog-doc#391

### DIFF
--- a/source/configuration/action/rsconf1_actionexeconlywhenpreviousissuspended.rst
+++ b/source/configuration/action/rsconf1_actionexeconlywhenpreviousissuspended.rst
@@ -1,15 +1,15 @@
 $ActionExecOnlyWhenPreviousIsSuspended
 --------------------------------------
 
-**Type:** global configuration directive
+**Type:** global configuration parameter
 
 **Default:** off
 
 **Description:**
 
-This directive allows to specify if actions should always be executed
+This parameter allows to specify if actions should always be executed
 ("off," the default) or only if the previous action is suspended ("on").
-This directive works hand-in-hand with the multiple actions per selector
+This parameter works hand-in-hand with the multiple actions per selector
 feature. It can be used, for example, to create rules that automatically
 switch destination servers or databases to a (set of) backup(s), if the
 primary server fails. Note that this feature depends on proper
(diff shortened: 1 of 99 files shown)


deoren:
@rgerhards Forgot to loop you in on this one.

rgerhards:
> Forgot to loop you in on this one.

haha, I get everything ;-) Actutally, I am subscribed to the repo at whole, so no need to mention me here explicitely (not that I would mind ;-)).

That said, I think the PR looks good. Any objections to merging it? As it fixes the trailing space issue, it is very likely to get merge conflicts if not merged soon.

rgerhards:
I forgot to mention that I see a lot of v1 files which we probably want to get rid of sooner or later. Just FYI...

deoren:
> @rgerhards: That said, I think the PR looks good. 

Thanks.

> @rgerhards: Any objections to merging it? As it fixes the trailing space issue, it is very likely to get merge conflicts if not merged soon.

Heh, looks like a conflict has already popped up. I'm not that familiar with resolving conflicts with Git, but I will take a look at that now.

deoren:
I think I have the changes integrated.

Steps I took:

1. `git fetch --all --prune` (download the latest changes from all remotes, toss branches deleted in other remotes)
1. `git checkout master`
1. `git merge upstream/master` (pull in the latest changes to the official master branch)
1. `git checkout  i-391-param-case` (the branch I'm using for this PR)
1. `git rebase master`
1.  At this point I manually modified the `source/configuration/modules/ompgsql.rst` file to remove the conflict markers and add in the original change for that file as part of this branch
1. `git rebase --continue`
1. `git push origin --force` (the assumption here is that I needed to overwrite my earlier work on the branch)

and here we are. Did I go about this the right way? I'm new to git, so any pointers or confirmation I'm not going about this the wrong way are appreciated.

deoren:
> @rgerhards: I forgot to mention that I see a lot of v1 files which we probably want to get rid of sooner or later. Just FYI...

Have those v1 files been superseded by newer files? In the past I landed on some of them when searching via Google and found the same content on newer pages. I don't recall if that was the case for all of the older v1 files.

deoren:
Updated title to reflect that we're moving forward with this.

rgerhards:
perfect rebase done :-)

deoren:
> perfect rebase done :-)

Nice, thanks for confirming! There is hope for me yet. :)

Issue #397, opened 2017-11-30
imuxsock module documentation: What exactly do the 'SysSock.Name' and 'SysSock.Use' parameters do?


deoren:
- Is the `SysSock.Name` parameter used to specify the replacement for `/dev/log` or is that parameter used to specify another socket that is _in addition to_ `/dev/log` as an input source?

- Does `SysSock.Use="on"` tell rsyslog to read from `/dev/log` or just from whatever is the default? If `SysSock.Name` is specified, does `SysSock.Use="on"` now refer to using the socket specified by `SysSock.Name`?

Refs: http://www.rsyslog.com/doc/v8-stable/configuration/modules/imuxsock.html

deoren:
Based on some testing, it appears that this configuration tells rsyslog to skip reading from `/dev/log` and read from `/run/systemd/journal/syslog` instead:

```
module(

    load="imuxsock"
 
    # Do not attempt to read /dev/log; systemd is controlling and supervising access to this socket
    sysSock.Use="off"

    # Here we replace what the stock /etc/rsyslog.d/list.conf fragment provides
    # by explicitly specifying the input socket name/path
    sysSock.Name="/run/systemd/journal/syslog"
)
```

deoren:
After much research (more to do), I believe I understand the relationship between the `/dev/log` socket, systemd and imuxsock a bit better, but I still have a few sticking points.

## Ubuntu

- `/dev/log` symlinked to `/run/systemd/journal/dev-log` (**write-target**)
-  imuxsock by default will read from `/run/systemd/journal/syslog` (**read-source**) as if you had specified 

## CentOS 7
- `/dev/log` a real socket, _owned_ by systemd, no symlink in place
- for RHEL-based systems, it appears you have to explicitly specify the `sysSock.Name` parameter as `sysSock.Name="/run/systemd/journal/syslog"` to receive the data that systemd passes over

deoren:
Actually, after some additional testing it appears that this will work fine for CentOS 7 systems:

```
module(load="imuxsock" sysSock.Use="off)
```

Is `sysSock.Use` used to control whether rsyslog attempts exclusive access to `/dev/log` (which conflicts with systemd owning that socket)?

deoren:
@rgerhards any pointers on this? After much testing I feel like I'm just going in circles. I've observed some behavior that sort of makes sense, but I'm not 100% sure I understand what is intended by the `SysSock.Use` and `SysSock.Name` parameters.

[...]

PR #427, opened 2017-12-12
imuxsock: update with details re systemd and systemd journal

rgerhards:
Behavior for systemd systems explicitely described after source
code review.

see also https://github.com/rsyslog/rsyslog-doc/issues/397

### DIFF
--- a/source/configuration/modules/imuxsock.rst
+++ b/source/configuration/modules/imuxsock.rst
@@ -62,6 +62,14 @@ properties at the end. For these reasons, the feature is **not enabled
 by default**. If you want to use it, you must turn it on (via
 SysSock.Annotate and Annotate).
 
+Note that on systems with **systemd journal**, the journal often interferes
+with the system log socket. Imuxsock can still be used in this setup
+and provides superior performance over :doc:`imjournal <imjournal>`, the
+alternative journal input module. It must be noted, however, that the
+journal tends to drop messages when it becomes busy instead of forwarding
+them to the system log socket. This is because the journal uses an async
+log socket interface for forwarding instead of the traditional synchronous one.
+
 Configuration Parameters
 ------------------------
 
@@ -83,6 +91,11 @@ Note: parameter names are case-insensitive.
    log socket. This is most useful if you run multiple instances of
    rsyslogd where only one shall handle the system log socket.
 -  **SysSock.Name** <name-of-socket>
+   This is the name of the system log socket, traditionally ``/dev/log``.
+   This socket is created upon rsyslog startup and deleted upon shutdown,
+   according to traditional syslogd behavior.
+
+   Please also be sure to read the special notes for systemd systems below!
 -  **SysSock.FlowControl** [on/**off**] - specifies if flow control
    should be applied to the system log socket.
 -  **SysSock.UsePIDFromSystem** [on/**off**] - specifies if the pid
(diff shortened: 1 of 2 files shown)


source/configuration/modules/imuxsock.rst
```diff
@@ -294,8 +334,8 @@ The following sample is used activate message annotation and thus
 trusted properties on the system log socket. module(load="imuxsock" #
 needs to be done just once SysSock.Annotate="on")
 
-Legacy Configuration Parameters
--------------------------------
+|FmtObsoleteDescription|
```
rgerhards: @deoren I did not have time to read up more intensely on RST: would it be possible to do a replacement of a complete text block? I mean we need to have the header and some introductory text be the same for all module descriptions. Would be great if we could get that from a single place...

rgerhards:
@deoren @davidelang I would appreciate if you could have a look at this PR. IMHO it describes everything that we know about the behavior.

source/configuration/modules/imuxsock.rst
deoren: @rgerhards I'm not sure if a complete block is supported, but I should be able to test sometime today to see if I can figure it out.

- http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#substitution-definitions
- http://docutils.sourceforge.net/docs/ref/rst/directives.html#include

Include statements are supported. Perhaps mixing the two would give the desired effect. 

That said, if it's the header syntax that you're having problems with, the number of characters beneath the text (or substitution reference) might be the issue. Literal text:

```
Legacy Configuration Parameters
-------------------------------
```

you would instead match the characters beneath to match the length of the substitution reference, not the text it will be expanded to when replaced:

```
|FmtObsoleteDescription|
------------------------
```

It took me a while to figure that out.

deoren: @rgerhards I've not found a way to bundle both the substitution text AND the header characters into a single substitution entry. Using an `include` file is the only way I know of to pull that off.

Is what you did with the substitution reference and header here a concern?

https://github.com/rgerhards/rsyslog-doc/commit/06b484a0e34549266ed02627acba9f8ee94a4050#diff-e80eae1b47ef8f45a205d11832bf4ffdR337

```
|FmtObsoleteDescription|
------------------------
```

That looks to be easy enough to maintain going forward. If we move the header level, we adjust the header characters out for the appropriate syntax. If we want to make sure that ALL module docs are consistent with header levels (something I was going to focus on after some other updates), then we could create an include file and use that, but wouldn't that be overkill here?

deoren:
@rgerhards Replied inline here and on the commit page itself (06b484a0e34549266ed02627acba9f8ee94a4050).

Pardon my lack of solid GitHub/Git experience, but if I had some recommended changes, how should I go about sharing them? Would I just clone your branch (I have your personal repo as a remote), make the changes, push a copy of my branch up to my repo and then link to the files here?

rgerhards:
> Pardon my lack of solid GitHub/Git experience, but if I had some recommended changes, how should I go about sharing them? Would I just clone your branch (I have your personal repo as a remote), make the changes, push a copy of my branch up to my repo and then link to the files here?

For pure text it's probably a bit messi... cloning the branch will work. Maybe you can add a correction PR and push this as PR. Then it should merge both (as it includes mine) when I merge yours. It just gets "interesting" if I want to add as well. Guess github was not really build for this. With source code the comments work well... ;-)

Another solution might be to really go ahead and merge and then do touchup PRs later (in the sense of let's do this as long as we really can't make it worse ;-).

deoren:
That makes sense. Fire when ready and I'll work from that point. 

Thank you for taking the time to review and modify the document, I know you have a lot of other items to also focus on.

rgerhards:
@deoren it's actually nice to work on this as a side activity. Also helps to keep your eyes open for potential trouble spots. And thanks for your really great help! It is most useful to have this kind of positive feedback and interest :-)

Issue #421, opened 2017-12-11
Update documentation to use substitution references in place of literal config format name text


deoren:
This is a continuation of the work started recently on #391. Now that the configuration format name substitutions have been defined, we need to update all references to the one of the three configuration format names to use them.

The goal is to make the names consistent going forward and (hopefully) allow for easier maintenance when/if the config format names are finalized (see #391 for a discussion of what those format names may be called in the near future).

I hope to submit a PR for this soon.

refs: #416, #414, #391

rgerhards:
as you go forward (and I can't today): would you mind shortening the name replacements? I thing the "...formatNew" (or so ;-)) part is not necessary. We should probably not need to refer to the old names more than once on the initial page. And the shorter replacements is more joy to type -- just my 2cts...

deoren:
> @rgerhards: as you go forward (and I can't today): would you mind shortening the name replacements?

My plan exactly.

> @rgerhards:  I thing the "...formatNew" (or so ;-)) part is not necessary.

Probably not. I think I was being overly verbose to keep it straight in my head while prototyping the setup.

> @rgerhards:  We should probably not need to refer to the old names more than once on the initial page. 

You're probably right. Whatever we do, I'll make sure to search the docs to make sure there is more than 1 use of each substitution. If not, I can prune any unused substitutions so that only active substitutions are kept.

> @rgerhards:  And the shorter replacements is more joy to type -- just my 2cts...

Understood. I have to admit that using editors like Microsoft Visual Code or Atom have spoiled me a bit with their offered suggestions as I type, so I may not be giving as much weight to shorter names as I should.

rgerhards:
I think this is closed. If not @deoren please re-open.

deoren:
@rgerhards looks like I don't have an option to open, just comment.

I was planning to use this issue to track the work of replacing the literal format names in the docs with the substitution references. I just haven't worked on this much yet due to my focus on other parts of the docs.

rgerhards:
> @rgerhards looks like I don't have an option to open, just comment.

ummm ... have you accepted the invitation for team membership?

deoren:
I haven't seen one. I will check my mail server to see if it has rejected mail from GitHub.

deoren:
@rgerhards if you have the option, could you resend the invite?

rgerhards:
@deoren I have updated it, not sure if that trigges a re-send...

deoren:
[...]

PR #428, opened 2017-12-12
Shorten format substitution reference names

deoren:
@rgerhards Since we're already using these references in other documents (e.g., #427), I'm going ahead and submitting this batch of changes now instead of waiting until the other docs have been updated.

refs rsyslog/rsyslog-doc#421

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -49,14 +49,9 @@ source_suffix = '.rst'
 rst_prolog = """
 .. |PRODUCT| replace:: ``Rsyslog``
 
-.. |FormatsBasicOldName| replace:: ``sysklogd``
-.. |FormatsBasicNewName| replace:: ``basic``
-
-.. |FormatsAdvancedOldName| replace:: ``RainerScript``
-.. |FormatsAdvancedNewName| replace:: ``advanced``
-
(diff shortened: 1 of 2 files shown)


source/conf.py
```diff
@@ -49,14 +49,9 @@
...
 
-.. |FormatsBasicOldName| replace:: ``sysklogd``
-.. |FormatsBasicNewName| replace:: ``basic``
-
-.. |FormatsAdvancedOldName| replace:: ``RainerScript``
-.. |FormatsAdvancedNewName| replace:: ``advanced``
-
-.. |FormatsObsoleteOldName| replace:: ``legacy``
-.. |FormatsObsoleteNewName| replace:: ``obsolete legacy``
+.. |FmtBasicNewName| replace:: ``basic``
```
rgerhards: sorry for being insistent... do we need "New" inside the name? ```FmtBasicName``` would be good IMHO...

deoren: No, we don't; good catch. I think I meant to do that based on your prior suggestion, but forgot about it when I made this quick change.

Provide via a separate commit or squash with the previous one?

rgerhards:
squash ;-)

deoren:
Corrected, squashed, changes pushed.

Issue #432, opened 2017-12-13
Are the imrelp's MaxDataSize and global() maxMessageSize parameters related?


deoren:
@rgerhards 

In the `source/rainerscript/global.rst` doc we have this:

> **`maxMessageSize`**
>
> The maximum message size rsyslog can process. Default is 8K. Anything above the maximum size will be truncated.

and in `source/configuration/modules/imrelp.rst` we have this:

> **``MaxDataSize <size_nbr>``**
>
> *Default is the global message size*
> 
>    Sets the max message size (in bytes) that can be received. Any messages above this size
>    will be rejected causing the relp client to reconnect and retry.

Are those two parameters related? It _looks_ like they are based on the "Default is the ..." remark, but I couldn't be sure (both because of the wording and because of what may be a bug). 

I'd like to update the `imrelp.rst` doc to be a little clearer on that point and also link to the `source/rainerscript/global.rst` for the current default value.

What is the intent of having the `MaxDataSize` imrelp parameter? Presumably to offer an override to the global setting, but is it a common need to have the values to be different? I'd like to note for that parameter a common use case of differing from the global setting.

Thanks.

Related:

- https://github.com/rsyslog/rsyslog-doc/blob/master/source/rainerscript/global.rst
- rsyslog/rsyslog#1782

rgerhards:
The doc tells what should happen, but as can be seen in the issue (and probably another one), there seems to be a bug in practice. Did not yet have time to look at it :-(

deoren:
> @rgerhards:  the doc tells what should happen

So if I understand what you're saying, then the `MaxDataSize` imrelp parameter should default to the global() `maxmessagesize` parameter if you do not explicitly set it.

What is the main reason for having the MaxDataSize imrelp parameter? Presumably to offer an override to the global setting, but is it a common need to have the values to be different? I'd like to note for that parameter a common use case of differing from the global setting.

rgerhards:
I barely remember there was a case, but essentially the sizes should be the same normally.

rgerhards:
actually (I am at the code now), imrelp uses the global default. Maybe the issue was only with omrelp.

deoren:
> @rgerhards actually (I am at the code now), imrelp uses the global default. 

Great, thanks for checking. I'll modify the wording so that the doc explicitly notes that.

> Maybe the issue was only with omrelp.

[...]

PR #433, opened 2017-12-13
Attempt to better emphasize relationship between global maxMessageSize, imrelp's MaxDataSize parameters

deoren:
@rgerhards Please let me know what you think.

refs rsyslog/rsyslog-doc#432

### DIFF
--- a/source/configuration/modules/imrelp.rst
+++ b/source/configuration/modules/imrelp.rst
@@ -57,7 +57,7 @@ Note: parameter names are case-insensitive.
 
 .. function:: MaxDataSize <size_nbr>
 
-   *Default is the global message size*
+   *Default is value of* :doc:`global(maxMessageSize) <../../rainerscript/global>` *parameter*
 
    Sets the max message size (in bytes) that can be received. Any messages above this size
    will be rejected causing the relp client to reconnect and retry.
--- a/source/rainerscript/global.rst
+++ b/source/rainerscript/global.rst
@@ -84,8 +84,11 @@ The following parameters can be set:
 
 - **maxMessageSize**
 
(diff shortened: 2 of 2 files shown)


deoren:
> @rgerhards: LGTM 

Thanks for the review. Sorry for the newbie question, but when merging I assume that I need to use the "Merge pull request" (aka, "Create a merge commit") option?

> @rgerhards: maybe we should add some info on the output modules? Even if we say its a bit uncertain?

Is there uncertainty about the intent or the results? 

If it's the results we're uncertain about, but you're confident that the intent is the same for both input and output modules, then I agree that we have enough information to make the changes to the docs. We state the intent as fact and then treat deviations as bugs to fix.

rgerhards:
> @rgerhards: LGTM
> Thanks for the review. Sorry for the newbie question, but when merging I assume that I need to use the "Merge pull request" (aka, "Create a merge commit") option?

yeah, that's the regular action. The other options are really for special cases. I have actually used "squash" only once (!) and the other never. You usually know when the merge doesn't sound good.

> 
> @rgerhards: maybe we should add some info on the output modules? Even if we say its a bit uncertain?
> Is there uncertainty about the intent or the results?
> 
> If it's the results we're uncertain about, but you're confident that the intent is the same for both input and output modules, then I agree that we have enough information to make the changes to the docs. We state the intent as fact and then treat deviations as bugs to fix.

Yeah, we can take that path. As I said, I would need to do a full code review to be 100% sure of the state, and I admit I am simple not able to do that right now.

deoren:
Added a note to #432 to create a separate PR for updating the output module docs.

PR #436, opened 2017-12-14
imuxsock: Additional details, section, alt wording

deoren:
Follow-up questions (for my next commit to the branch behind this PR):

- **TODO**: How does rsyslog obtain the log socket from systemd?

- **TODO**: Is ``sysSock.Use`` ignored for systemd systems? It seems to be (I saw the statement that the `SysSock.` parameters are often ignored).

refs rsyslog/rsyslog-doc#397

**EDIT**

The branch sourced by this PR is viewable here: 

http://rsyslog.whyaskwhy.org/i-397/configuration/modules/imuxsock.html

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -30,7 +30,7 @@ sys.path.insert(0, os.path.abspath('_ext'))
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['edit_on_github']
+extensions = ['edit_on_github', 'sphinx.ext.todo']
 edit_on_github_project = 'rsyslog/rsyslog-doc'
 edit_on_github_branch = 'master'
 
@@ -136,6 +136,13 @@ pygments_style = 'sphinx'
 
 # Warn about all references where the target cannot be found. Default is False.
 nitpicky = True
+
+# If true, `todo` and `todoList` produce output, else they produce nothing.
+todo_include_todos = True
+
+# If this is True, todo emits a warning for each TODO entries. The default
+# is False.
+todo_emit_warnings = True
 
 # -- Options for HTML output ---------------------------------------------------
 
--- a/source/configuration/modules/imuxsock.rst
+++ b/source/configuration/modules/imuxsock.rst
@@ -1,34 +1,244 @@
 imuxsock: Unix Socket Input
-===========================
+***************************
 
 ===========================  ===========================================================================
 **Module Name:**             **imuxsock**
 **Author:**                  `Rainer Gerhards <http://www.gerhards.net/rainer>`_ <rgerhards@adiscon.com>
 ===========================  ===========================================================================
 
-**Provides the ability to accept syslog messages via local Unix sockets.
-Most importantly, this is the mechanism by which the syslog(3) call
-delivers syslog messages to rsyslogd.** So you need to have this module
-loaded to read the system log socket and be able to process log messages
-from applications running on the local system.
 
-**Application-provided timestamps are ignored by default.** This is
-needed, as some programs (e.g. sshd) log with inconsistent timezone
-information, what messes up the local logs (which by default don't even
-contain time zone information). This seems to be consistent with what
-sysklogd did for the past four years. Alternate behaviour may be
-desirable if gateway-like processes send messages via the local log slot
-- in this case, it can be enabled via the IgnoreTimestamp and
-SysSock.IgnoreTimestamp config directives
+Purpose
+=======
 
-**There is input rate limiting available,** (since 5.7.1) to guard you
-against the problems of a wild running logging process. If more than
-SysSock.RateLimit.Interval \* SysSock.RateLimit.Burst log messages are
-emitted from the same process, those messages with
-SysSock.RateLimit.Severity or lower will be dropped. It is not possible
+This module provides the ability to accept syslog messages from applications
+running on the local system via Unix sockets. Most importantly, this is the
+mechanism by which the :manpage:`syslog(3)` call delivers syslog messages
+to rsyslogd.
+
+Notable Features
+================
+
+- :ref:`rate-limiting-label`
+- :ref:`trusted-properties-label`
+- :ref:`flow-control-label`
+- :ref:`application-timestamps-label`
+- :ref:`systemd-details-label`
+
+
+Configuration Parameters
+========================
+
+.. note::
+
+   Parameter names are case-insensitive.
+
+Global Parameters
+-----------------
+
+.. warning::
+
+   When running under systemd, **many "sysSock." parameters are ignored**.
+   See parameter descriptions and the :ref:`systemd-details-label` section for
+   details.
+
+-  **SysSock.IgnoreTimestamp** [**on**/off]
+   Ignore timestamps included in the messages, applies to messages
+   received via the system log socket.
+
+-  **SysSock.IgnoreOwnMessages** [**on**/off] (available since 7.3.7)
+   Ignores messages that originated from the same instance of rsyslogd.
+   There usually is no reason to receive messages from ourselves. This
+   setting is vital when writing messages to the Linux journal. See
+   :doc:`omjournal <omjournal>` module documentation for a more
+   in-depth description.
+
+-  **SysSock.Use** (imuxsock) [**on**/off] - Listen on the default local
+   log socket (``/dev/log``) or, if provided, use the log socket value
+   assigned to the ``SysSock.Name`` parameter instead of the default. This
+   is most useful if you run multiple instances of rsyslogd where only one
+   shall handle the system log socket.  Unless disabled by the
+   ``SysSock.Unlink`` setting, this socket is created upon rsyslog startup
+   and deleted upon shutdown, according to traditional syslogd behavior.
+
+   The behavior of this parameter is different for systemd systems. For those
+   systems, ``SysSock.Use`` still needs to be enabled, but the value of
+   ``SysSock.Name`` is ignored and the socket provided by systemd is used
+   instead. If this parameter is *not* enabled, then imuxsock will only be
+   of use if a custom input is configured.
+
+   See the :ref:`systemd-details-label` section for details.
+
+-  **SysSock.Name** <name-of-socket> - specifies an alternate log socket
+   to be used instead of the default system log socket, traditionally
+   ``/dev/log``. Unless disabled by the ``SysSock.Unlink`` setting,
+   this socket is created upon rsyslog startup and deleted upon shutdown,
+   according to traditional syslogd behavior.
+
+   The behavior of this parameter is different for systemd systems. See the
+   the :ref:`systemd-details-label` section for details.
+
+-  **SysSock.FlowControl** [on/**off**] - specifies if flow control
+   should be applied to the system log socket.
+
+-  **SysSock.UsePIDFromSystem** [on/**off**] - specifies if the pid
+   being logged shall be obtained from the log socket itself. If so, the
+   TAG part of the message is rewritten. It is recommended to turn this
+   option on, but the default is "off" to keep compatible with earlier
+   versions of rsyslog.
+
+-  **SysSock.RateLimit.Interval** [number] - specifies the rate-limiting
+   interval in seconds. Default value is 0, which turns off rate
+   limiting. Set it to a number of seconds (5 recommended) to activate
+   rate-limiting. The default of 0 has been chosen as people experienced
+   problems with this feature activated by default. Now it needs an
+   explicit opt-in by setting this parameter.
+
+-  **SysSock.RateLimit.Burst** [number] - specifies the rate-limiting
+   burst in number of messages. Default is 200.
+
+-  **SysSock.RateLimit.Severity** [numerical severity] - specifies the
+   severity of messages that shall be rate-limited.
+
+-  **SysSock.UseSysTimeStamp** [**on**/off] the same as the input
+   parameter ``UseSysTimeStamp``, but for the system log socket. See
+   description there.
+
+-  **SysSock.Annotate** <on/**off**> turn on annotation/trusted
+   properties for the system log socket.
+
+-  **SysSock.ParseTrusted** <on/**off**> if Annotation is turned on,
+   create JSON/lumberjack properties out of the trusted properties
+   (which can be accessed via RainerScript JSON Variables, e.g. ``$!pid``)
+   instead of adding them to the message.
+
+-  **SysSock.Unlink** <**on**/off> (available since 7.3.9)
+   if turned on (default), the system socket is unlinked and re-created
+   when opened and also unlinked when finally closed. Note that this
+   setting has no effect when running under systemd control (because
+   systemd handles the socket).
+
+-  **sysSock.useSpecialParser** (available since 8.9.0)
+   The equivalent of the ``useSpecialParser`` input parameter for the
+   system socket.
+
+-  **sysSock.parseHostname** (available since 8.9.0)
+   The equivalent of the ``parseHostname`` input parameter for the
+   system socket.
+
+Input Parameters
+----------------
+
+-  **ruleset** [name]
+   Binds specified ruleset to this input. If not set, the default
+   ruleset is bound. (available since 8.17.0)
+
+-  **IgnoreTimestamp** [**on**/off]
+   Ignore timestamps included in messages received from the input being
+   defined.
+
+-  **IgnoreOwnMessages** [**on**/off] (available since 7.3.7)
+   Ignore messages that originated from the same instance of rsyslogd.
+   There usually is no reason to receive messages from ourselves. This
+   setting is vital when writing messages to the Linux journal. See
+   :doc:`omjournal <omjournal>` module documentation for a more
+   in-depth description.
+
+-  **FlowControl** [on/**off**] - specifies if flow control should be
+   applied to the input being defined.
+
+-  **RateLimit.Interval** [number] - specifies the rate-limiting
+   interval in seconds. Default value is 0, which turns off rate
+   limiting. Set it to a number of seconds (5 recommended) to activate
+   rate-limiting. The default of 0 has been chosen as people experienced
+   problems with this feature activated by default. Now it needs an
+   explicit opt-in by setting this parameter.
+
+-  **RateLimit.Burst** [number] - specifies the rate-limiting burst in
+   number of messages. Default is 200.
+
+-  **RateLimit.Severity** [numerical severity] - specifies the severity
+   of messages that shall be rate-limited.
+
+-  **UsePIDFromSystem** [on/**off**] - specifies if the pid being logged
+   shall be obtained from the log socket itself. If so, the TAG part of
+   the message is rewritten. It is recommended to turn this option on,
+   but the default is "off" to keep compatible with earlier versions of
+   rsyslog.
+
+-  **UseSysTimeStamp** [**on**/off] instructs imuxsock to obtain message
+   time from the system (via control messages) instead of using time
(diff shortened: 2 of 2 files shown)


rgerhards:
> TODO: How does rsyslog obtain the log socket from systemd?

via a system call to systemd -- I don't think the specifics need to go into the user doc. 
> 
> TODO: Is sysSock.Use ignored for systemd systems? It seems to be (I saw the statement that the SysSock. parameters are often ignored).

Need to verify once again, but I think it depends on configuration. If systemd uses socket activation and journald forward to the socket, it possibly is ignored. If either is not the case, it is not ignored. But it's more a systemd question than a rsyslog one...

deoren:
@rgerhards I've done some additional work on this doc. In the last few commits I tested the "todo", "note" and "warning" directives to see how calling out the specific points would help.

I also enabled an option to treat TODO entries as warnings. This results in Travis CI encountering those warnings and considering the build as failed. 

I consider this a useful effect: 

1. branch from master
1. start work and add various TODO entries for incomplete tasks
1. Travis CI flags that work as part of a failed build
1. Finish work, remove TODO entries
1. Travis CI marks the build as a pass

This practice could help the work on a PR as not quite ready to merge. I'll probably use the habit going forward (the intent being to remove all todo entries before merge).

Regarding the final TODO item, I browsed through the source, but I can't be confident that I looked over the right spots to confirm whether `sysSock.Use` is used when systemd support is enabled.

1. [Line 1440](https://github.com/rsyslog/rsyslog/blob/ac87ab0a71d99b0093be4bf159c3f119454fb442/plugins/imuxsock/imuxsock.c#L1440) - the local log socket index starts at 0 provided that `sysSock.Use="off"` is not set (aka, `$OmitLocalLogging on`)
1. [Line 1138](https://github.com/rsyslog/rsyslog/blob/ac87ab0a71d99b0093be4bf159c3f119454fb442/plugins/imuxsock/imuxsock.c#L1138) - the index position 0 local socket's sockName value is set to `/dev/log`
1. If the `LogSockName` config object parameter is not NULL then it is used
1. If the `LogSockName` config object parameter _is_ NULL then a systemd boot check is performed
1. If all systemd related checks pass then index position 0 local socket's sockName value is set to `/run/systemd/journal/syslog` which overwrites the earlier `/dev/log` value

I don't understand the code well enough to tell if `sysSock.Name` is set to `/dev/log` if not explicitly specified _prior_ to the check on line 1143.

deoren:
@rgerhards Any thoughts on the items noted in https://github.com/rsyslog/rsyslog-doc/pull/436#issuecomment-352167572?

rgerhards:
@deoren sorry, will check ASAP. But I guess my findings will be the same like in the last time (nevertheless a re-check is worth doing). Many things depend on systemd configuration...

rgerhards:
> I consider this a useful effect:
> 
> branch from master
> start work and add various TODO entries for incomplete tasks
> Travis CI flags that work as part of a failed build
> Finish work, remove TODO entries
> Travis CI marks the build as a pass
> This practice could help the work on a PR as not quite ready to merge. I'll probably use the habit going forward (the intent being to remove all todo entries before merge).

I would suggest adding this to README or CONTRIBUTING doc, it's very useful info.

> Line 1440 - the local log socket index starts at 0 provided that sysSock.Use="off" is not set (aka, $OmitLocalLogging on)
yes
> Line 1138 - the index position 0 local socket's sockName value is set to /dev/log
yes
> If the LogSockName config object parameter is not NULL then it is used
yes
> If the LogSockName config object parameter is NULL then a systemd boot check is performed
yes
> If all systemd related checks pass then index position 0 local socket's sockName value is set to /run/systemd/journal/syslog which overwrites the earlier /dev/log value
yes
> I don't understand the code well enough to tell if sysSock.Name is set to /dev/log if not explicitly specified prior to the check on line 1143.
no, I don't see that. Even if it were, it is always set in line 1140, overwriting whatever would potentially have been stored before.

...more follows...

deoren:
@rgerhards Thanks for bearing with me as a beat a dead horse here. 

If I understand everything that has been said, then this would be the equivlanet pseudo-code  for [`imuxsock.c` as of commit ac87ab0a71d99b0093be4bf159c3f119454fb442](https://github.com/rsyslog/rsyslog/blob/ac87ab0a71d99b0093be4bf159c3f119454fb442/plugins/imuxsock/imuxsock.c#L1138):

```
if (the user hasn't explicitly chosen to set SysSock.Use="off") {

    /* system socket */
    listeners[0].sockName = "/dev/log";

    if (the user has specified sysSock.Name) {
        then go ahead and use that.
    }

    else if (systemd is present & running) {
        if (the systemd provided socket exists and is a valid socket) {
            listen on the socket
        }
    }
}
```

If so, then this seems to imply that `SysSock.Use` has to be set to `SysSock.Use="on"`, either by the default setting or explicitly set, before the systemd socket will be used. In other words, instead of this being a question of whether the `SysSock.Use` setting is honored as I mused before, that parameter being set to `on` appears to be *required* before systemd support is used.

deoren:
@rgerhards I've worked on the doc some more, using what I _believe_ I know to rework some of my earlier modifications to the doc. I've also made multiple other changes while working on the doc which were not part of the scope of my original work.

> > @deoren: I consider this a useful effect: 
> > 
> > 1. branch from master
> > 1. start work and add various TODO entries for incomplete tasks
> > 1. Travis CI flags that work as part of a failed build
> > 1. Finish work, remove TODO entries
> > 1. Travis CI marks the build as a pass
> > 
> > This practice could help the work on a PR as not quite ready to merge. I'll probably use the habit going forward (the intent being to remove all todo entries before merge).
>
> @rgerhards: I would suggest adding this to README or CONTRIBUTING doc, it's very useful info.

Sure. I opted to try it with the work on this PR to see how well you would receive the idea. Since you approve, I'll definitely make plans to cover it in a future CONTRIBUTING doc. I'll move that up my list as I think it would be useful to have at least a skeleton doc in place to reference as I continue working on the other content.

On the note of unplanned tweaks, I went ahead and dropped all obsolete legacy coverage from this doc. Am I right in understanding that this is our goal for future doc edits? We go to modify the doc for another purpose and while we're there, we go ahead and prune the obsolete legacy coverage?

What do you think about the doc as it stands now (db1f1618ac2577c06e4eea778a8c1054f29e2430)?

rgerhards:
sorry for yet another partial answer

> On the note of unplanned tweaks, I went ahead and dropped all obsolete legacy coverage from this doc. Am I right in understanding that this is our goal for future doc edits? We go to modify the doc for another purpose and while we're there, we go ahead and prune the obsolete legacy coverage?

I think this is the best path forward. I would also prune any remaining copyright notices on individual pages. They stem back to the original hand-crafted html doc and are no longer needed as sphinx generates the footer. Just nobody removed them so far, but it is fine to do so.

deoren:
> @rgerhards: sorry for yet another partial answer

I forgive you, provided that you're only dodging the questions due to the amount of time necessary to cover everything. ;)

In general, what do you think about the doc as it stands now (db1f161)?

> > @deoren: On the note of unplanned tweaks, I went ahead and dropped all obsolete legacy coverage from this doc. Am I right in understanding that this is our goal for future doc edits? We go to modify the doc for another purpose and while we're there, we go ahead and prune the obsolete legacy coverage?
>
> @rgerhards: I think this is the best path forward. 

I'll adopt that approach then.

> @rgerhards: I would also prune any remaining copyright notices on individual pages. They stem back to the original hand-crafted html doc and are no longer needed as sphinx generates the footer. Just nobody removed them so far, but it is fine to do so.

I'll try to tackle that on #455 and will also add that as an item to a guidelines/contributing doc that I'm drafting (#451).

rgerhards:
> @rgerhards: sorry for yet another partial answer
> 
> I forgive you, provided that you're only dodging the questions due to the amount of time necessary to cover everything. ;)

hehe  - yeah ```imuxsock``` is becoming my favorite troublemaker ;-) Actually, it has "evolved" quite a bit and it got less and less attention with the systemd journal actions...

OK, one thing I was puzzled about was the processing in ```openLogsocket()``` here: https://github.com/rsyslog/rsyslog/blob/ac87ab0a71d99b0093be4bf159c3f119454fb442/plugins/imuxsock/imuxsock.c#L541 But I double-checked and this does not interfere with the system log socket. It just enables sockets to be passed in via systemd socket activation. In that case, the socket path name must also match. So while this affects the system log socket, it does *not* overwrite it's name contrary to what I originally thought. 

A notable effect of that check is: **iff** the socket name is passed in via systemd (socket activation), the socket is not touched. Otherwise, it is created on imuxsock startup and removed upon shutdown (traditional syslog behavior). Note especially that this applies to the system log socket as well: no matter what the name is set to (and how this happened), it is just the socket activation check that decides over creation and deletion of the socket file system entry (well, actually the ```unlink``` config parameter also controls destruction on an per-input layer).

Sorry, it's all pretty complex and partly inconsistent and far from intuitive. I guess there is a little bit too much legacy inside this module... For the same reason, I wouldn't mind if someone tripple checks, but what is given here is really to best of my knowledge and after the 4th review (I think) -- shame that I also see new things each time I look...

The bottom line IMHO is that the proposed description is right, but could be amended with the info given in this post. Maybe we can manage to structure it in a way that folks have a chance to understand...

deoren:
> @rgerhards OK, one thing I was puzzled about was the processing in openLogsocket() here: https://github.com/rsyslog/rsyslog/blob/ac87ab0a71d99b0093be4bf159c3f119454fb442/plugins/imuxsock/imuxsock.c#L541 But I double-checked and this does not interfere with the system log socket. It just enables sockets to be passed in via systemd socket activation. In that case, the socket path name must also match. So while this affects the system log socket, it does not overwrite it's name contrary to what I originally thought.

What is the order/relation between activateListeners() and openLogSocket()? I think I read the code wrong and assumed incorrectly that activateListeners() was called before openLogSocket(). 

> @rgerhards: A notable effect of that check is: iff the socket name is passed in via systemd (socket activation), the socket is not touched. Otherwise, it is created on imuxsock startup and removed upon shutdown (traditional syslog behavior). Note especially that this applies to the system log socket as well: no matter what the name is set to (and how this happened), it is just the socket activation check that decides over creation and deletion of the socket file system entry (well, actually the unlink config parameter also controls destruction on an per-input layer).

If I interpreted the order correctly, wouldn't a custom setting for the `SysSock.Name` parameter (_and_ if systemd support is enabled/available) result in rsyslog not listening on the `/run/systemd/journal/syslog` socket? On the other hand, if `SysSock.Name` is not set to a custom value, wouldn't `/dev/log` be ignored entirely: not read, not unlinked at daemon start or stop?

deoren:
> @rgerhards: Maybe we can manage to structure it in a way that folks have a chance to understand...

I was thinking on that when I last modified the doc. In my edits I took and created a subsection just for the systemd content right where you introduced the new material. I then added a brief description of how the the `SysSock.Use` and `SysSock.Name` parameters are affected by a working systemd setup and then pointed the reader to the new section.

It may not be an ideal approach, but it feels like a workable one as a first/second pass. I also noticed that the Annotation parameters were referenced as if they were new, but without a recommendation to use JSON message properties to avoid modifying the original message. That paragraph could probably use updating as part of some later work to refresh the whole doc. I mention that block just as a way of saying that we can loop back and update the doc further once we get the core imuxsock/systemd integration info into the doc in a reasonably clear manner.

rgerhards:
> What is the order/relation between activateListeners() and openLogSocket()? I think I read the code wrong and assumed incorrectly that activateListeners() was called before openLogSocket().

activateListeners() *calls* openLogSocket(), so you actually read it right, activateListeners() is called first,. Basically it does:

* handle the system log socket setup (iff useSyssock="on")
* then iterates through all configured sockets (including system log socket) and calls openLogSocket() to open/aquire them (aquire happens if systemd passes sockets in via socket activation configuration).



> If I interpreted the order correctly, wouldn't a custom setting for the SysSock.Name parameter (and if systemd support is enabled/available) result in rsyslog not listening on the /run/systemd/journal/syslog socket? 

this is right. It would listen on the configured socket name insteand.

> On the other hand, if SysSock.Name is not set to a custom value, wouldn't /dev/log be ignored entirely: not read, not unlinked at daemon start or stop?

prequisites:
* systemd present, 
* /run/systemd/journal/syslog socket NOT existing

The it listens on /dev/log, because this is the default value (https://github.com/rsyslog/rsyslog/blob/ac87ab0a71d99b0093be4bf159c3f119454fb442/plugins/imuxsock/imuxsock.c#L1138) and that default is not overwritten.

It is created/unlinked iff:
* systemd is configured to use socket activation
* socket activation is active on "/dev/log"

deoren:
> @rgerhards: prequisites:
>
> 
> - systemd present,
> - /run/systemd/journal/syslog socket NOT existing

Ah ok, line 1147 through 1149: the default is only overwritten if systemd is present and the socket exists. Otherwise, I see what you're saying, the default is kept (because `SysSock.Name` is also not set in this scenario) and openLogSocket is called against it.

> @rgerhards: The it listens on /dev/log, because this is the default value (https://github.com/rsyslog/rsyslog/blob/ac87ab0a71d99b0093be4bf159c3f119454fb442/plugins/imuxsock/imuxsock.c#L1138) and that default is not overwritten.

I think I understand now.

> It is created/unlinked iff:
> 
> - systemd is configured to use socket activation
> - socket activation is active on "/dev/log"

I found this page:

http://0pointer.de/blog/projects/socket-activation.html

which makes sense. I think once I look over it again in more depth I will better understand it.

I'll review the doc again when I get some time and try to make sure that what it states correctly portrays what you've noted here.

Thanks again for the details.

rgerhards:
@deoren thanks for **your** persistence -- it's so ugly, without a bit of being pushed I'd probably not investigated that any further ;-)

deoren:
> @rgerhards: @deoren thanks for your persistence -- it's so ugly, without a bit of being pushed I'd probably not investigated that any further ;-)

You're quite welcome. I appreciate your patience as we step through this. I think I understand the situation now, but I'd like to get your feedback on this to make sure:

Updated pseudo-code:

```c
if (the user has not explicitly chosen to set SysSock.Use="off") {

    set the default listener socket name to "/dev/log"

    if (the user has specified sysSock.Name="/path/to/custom/socket") {
        listen on the user-specified socket
        do not manage (unlink, create, unlink later) the "/dev/log" socket
        do not listen on the "/dev/log" socket
        do not listen on the systemd-provided socket
    }

    else if (systemd is present and running) {

        if (the systemd-provided socket exists and is a valid socket) {
            listen on the systemd-provided socket
        do not manage (unlink, create, unlink later) the "/dev/log" socket
        do not listen on the "/dev/log" socket
        }

        else if (the socket does not exist) {
            unlink "/dev/log"
            recreate "/dev/log"
            listen on the "/dev/log" socket
            unlink "/dev/log" socket again during shutdown phase
        }

    }

}
```

For the situation where all of these are true:

- `SysSock.Use="on"` (by default or explicitly set)
- systemd is present
- the `/run/systemd/journal/syslog` socket does not exist
- a custom value has not been provided for `SysSock.Name`

then rsyslog falls back to "owning" `/dev/log` by unlinking it, recreating it and then taking responsibility for unlinking it again upon shutdown (with default `SysSock.Unlink` value).

I assume that this is not treated as an error condition because the user could have intentionally configured their system this way? Perhaps as a means of bypassing systemd for performance reasons as I've seen you and @davidelang mention before?

rgerhards:
just a quick status update: it's not 100% right. The name selection and create/remove are untangled. I'll see that I update the pseudocode, it's a very good start ... but It may become as large as the C code itself - simply because it is complicated...

deoren:
> @rgerhards: just a quick status update: it's not 100% right

Bummer, I thought I had the gist of it figured out.

> @rgerhards: The name selection and create/remove are untangled. I'll see that I update the pseudocode

Thank you!

> @rgerhards: it's a very good start ... but It may become as large as the C code itself - simply because it is complicated...

Understood. We can always trim if the results are too massive for consumption, so there is that. Detailed now, revise doc to capture the important bits and then decide the next step.

rgerhards:
This should it be:
```
FIRST STEP: select NAME of system socket:

if (the user has not explicitly chosen to set SysSock.Use="off") {
    set the default listener socket name to "/dev/log"
    if (the user has specified sysSock.Name="/path/to/custom/socket") {
    	overwrite socket name with user-specified socket
    } else if  (running under sytemd and /run/systemd/journal/syslog exists)
    	overwrite socket name with "/run/systemd/journal/syslog"
}
END FIRST STEP: name is set now

SECOND STEP: listen on socket
Note: this is true for all sockets, be it system socket or not. But if
SysSock.Use="off", the system socket will not be listened on.

    forall sockets to be activated {
    	if socket is passed in via systemd (name is checked) {
		use as sockets as is, do not create or unlink
	} else {
		create log socket, open it
	}
    }

 
THIRD STEP: shutdown
    forall sockets to be shut down {
    	if socket was passed in via systemd (name is checked) {
		do nothing
	} else if (unlink is permitted by configuration param{
		unlink log socket
	} else {
		do nothing
	}
    }
```

Note the three independent steps.

rgerhards:
I did not explicitely mention it: I think if you try to mix the three steps, you end up with many more possibilities, none of which are really logically related. Mabye we need this for the doc. In that case, probably a big table-like structure (or bullet points) may be best...

deoren:
Thanks for the feedback. I will try to work it into a draft for review, but it may take me some time to do so (busy week).

deoren:
Nothing recent to show on this, but the current branch is viewable here:

http://rsyslog.whyaskwhy.org/i-397/configuration/modules/imuxsock.html

rgerhards:
@deoren What about if we merge as-is and do further improvements as we go by? I think the current PR **is** an improvement...

deoren:
@rgerhards I have a little more work to do on it and hope to have it ready to demo soon. I really like the detail you provided in your recent post on this PR and hope to incorporate that in before the merge. I also plan to squash the commits to make it cleaner.

deoren:
>@deoren: @rgerhards I have a little more work to do on it and hope to have it ready to demo soon. I really like the detail you provided in your recent post on this PR and hope to incorporate that in before the merge. I also plan to squash the commits to make it cleaner.

I've got the bulk of the content in and absent the cleanup work necessary to make sure I haven't forgotten anything obvious, I think this branch is in RC status for merging.

- Latest master branch: http://rsyslog.whyaskwhy.org/master/configuration/modules/imuxsock.html
- Current test branch: http://rsyslog.whyaskwhy.org/i-397/configuration/modules/imuxsock.html

Anything leap out at you as glaringly wrong/awful?

rgerhards:
@deoren I leave the final judgment to you, but I would suggest to merge.

deoren:
> @rgerhards: Looks good to me. 

Thanks for looking over it.

> @rgerhards: There are some rough edges

Definitely. I've already got a mental list of some further refinements to do as part of another PR.

> @rgerhards: everything is correct. IMHO we should merge now and work on improvements later on. The current state is a huge improvement....

Thanks, and yes I agree. Assuming it's preferable, let me squash the commits to reduce the log chatter. I'll then merge the results and close out the related ticket.

rgerhards:
squash is a good idea - go, go, go :-)

deoren:
> @rgerhards: squash is a good idea - go, go, go :-)

Squashed, pushed and waiting for CI build to complete.

deoren:
Looks like Travis had some recent issues. I assume that the delay in building is related to that. I've built on my VPS and locally without issue, so I'm going to go with those results. Merging.

rgerhards:
Just FYI: travis seems to be overloaded the past 2 weeks. I see long delays / fail due to timeout frequently. Either the OSS community is ultra-busy around the vacation period or the Travis folks have a very good vacation ;-)

deoren:
@rgerhards Maybe both. :)

PR #437, opened 2017-12-15
imfile: Add clarification re state files, fix typos

deoren:
- Fix typos
- Emphasize that unless the PersistStateInterval parameter is set,
  no state files will be created
- Note that state files are only read on startup
- Mention "advanced format" global workDirectory parameter alongside
  the $WorkDirectory directive for comparison

refs rsyslog/rsyslog#2159

### DIFF
--- a/source/configuration/modules/imfile.rst
+++ b/source/configuration/modules/imfile.rst
@@ -70,16 +70,13 @@ State Files
 ...........
 Rsyslog must keep track of which parts of the monitored file
 are already processed. This is done in so-called "state files" that
-are created in the rsyslog working directory. These files are read
-on startup to resume monitoring after a shutdown.
+are created in the rsyslog working directory and are read on startup to
+resume monitoring after a shutdown. The location of the rsyslog
+working directory is configurable via the ``global(workDirectory)``
+|FmtAdvancedName| format parameter.
 
 **Note**: The ``PersistStateInterval`` parameter must be set, otherwise state
 files will NOT be created.
-
-The location of the rsyslog working directory is configurable via the
-``$WorkDirectory`` |FmtObsoleteName| format (limited to |FmtObsoleteName|
-format directives only) or the ``global(workDirectory)`` |FmtAdvancedName|
-format parameter (|FmtAdvancedName| format parameters only).
 
 To avoid problems with duplicate state files, rsyslog automatically
 generates state file names according to the following scheme:
--- a/source/configuration/modules/imfile.rst
+++ b/source/configuration/modules/imfile.rst
@@ -399,17 +399,6 @@ Note: parameter names are case-insensitive.
    This is used to turn on or off the addition of the "@cee:" cookie to the
    message object.
 
-.. function:: stateFile [name-of-state-file]
-
-   **Default: unset**
-
-   **This parameter is deprecated.** It still is accepted, but should
-   no longer be used for newly created configurations.
-
-   This is the name of this file's state file. This parameter should
-   usually **not** be used. Check the section on "State Files" above
-   for more details.
-
 .. index::
    single: imfile; reopenOnTruncate
 .. function:: reopenOnTruncate [on/off] (requires v8.16.0+)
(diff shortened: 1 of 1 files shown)


source/configuration/modules/imfile.rst
```diff
@@ -69,9 +69,17 @@ disabled by default. It can explicitly be turned on or off via the
...
-(configurable via $WorkDirectory).
+are already processed. This is done in so-called "state files" that
+are created in the rsyslog working directory. These files are read
+on startup to resume monitoring after a shutdown.
+
+**Note**: The ``PersistStateInterval`` parameter must be set, otherwise state
+files will NOT be created.
+
+The location of the rsyslog working directory is configurable via the
+``$WorkDirectory`` |FmtObsoleteName| format (limited to |FmtObsoleteName|
```
rgerhards: I would completely drop the obsolete format. That only encourages use. Let's get rid of it everywhere except in the short cross-ref sections at the end of the config objects (e.g,. plugin like imuxsock).

source/configuration/modules/imfile.rst
```diff
@@ -396,7 +406,7 @@ Note: parameter names are case-insensitive.
 
    **Default: unset**
 
-   **This paramater is deprecated.** It still is accepted, but should
+   **This parameter is deprecated.** It still is accepted, but should
```
rgerhards: We probably should warn against the use. A common error in those days where it was necessary was that folks copy-and-pasted the def, ending up with dupe state file names, which caused (and I thinks still causes) rsyslog to get totally confused. To avoid this, we now have the automatic name generation.

Maybe an extra section at the end for deprecated attributes?

source/configuration/modules/imfile.rst
deoren: Fair enough. I was just attempting to balance the mention of the obsolete legacy format directive, thinking that we could loop back later as part of a focused effort to prune legacy directives.

However I think it makes good sense to replace use of the old directives with the new/advanced format equivalents as we find them.

source/configuration/modules/imfile.rst
deoren: > Maybe an extra section at the end for deprecated attributes?

That makes sense. Here though, I was just fixing a typo that my editor flagged.

rgerhards: yeah, I used the opportunity ;-)

source/configuration/modules/imfile.rst
rgerhards: IMHO the obsolete statements inside doc text, tutorials etc only are there because nobody yet too the time to change them to advanced format. From the mailing list and issues I know that many people find it confusing to see this mix of obsolete and advanced statements (and I think they are right with that).

The cure actually is to change the statement and not mention the obsolete ones. I still think we should provide a very brief reference of old-to-new mapping, but not more than that. Whoever wants more info on the old statements should refer to old doc. One reason is that it often is really very hard to accurately describe what obsolete format does, because it is so counter-intuitive and context-depending (from a CS point of view, it's actually a context depending mes^h^h^hlanguage).

deoren: > @rgerhards: IMHO the obsolete statements inside doc text, tutorials etc only are there because nobody yet took the time to change them to advanced format.

That and perhaps there was a question of whether the doc should include both or only the new format. I think we've clarified some of that with recent discussions, though it may be a good idea to create a separate issue that makes that clear and could serve as a good pointer if/when questions about the work comes up.

source/configuration/modules/imfile.rst
```diff
@@ -605,8 +601,8 @@ Legacy Configuration Directives
 
    Equivalent to: Ruleset
 
-Legacy Example
-^^^^^^^^^^^^^^
```
rgerhards: I think it is too early now, but  in the end we should also remove obsolete examples. We need new style instead ;-)

source/configuration/modules/imfile.rst
deoren: > @rgerhards: The cure actually is to change the statement and not mention the obsolete ones. I still think we should provide a very brief reference of old-to-new mapping, but not more than that. 

Do you see that goal being accomplished by way of the "obsolete format directives" section at the bottom of module pages?

source/configuration/modules/imfile.rst
deoren: That makes sense. I'll open a new issue soon and will try to include some targeted questions to make sure I have the right focus as I modify the docs going forward.

rgerhards:
> @rgerhards Added two new commits. I've not reviewed the header levels in the doc to make sure they match the spec as I hope to do that as part of a later effort (where I'll focus just on header levels).

there exist some closed-unmerged PRs from @mostolog which were unfortunately unfinished and thus not mergable. I think one of them addressed the header levels. You might want to search through the closed PRs -- it's not that many.

deoren:
> there exist some closed-unmerged PRs from @mostolog which were unfortunately unfinished and thus not mergable. I think one of them addressed the header levels. You might want to search through the closed PRs -- it's not that many.

Will do. 

What do you think about the state of the doc as of 092951221593bf117f73c0500c55e50951d82053? Anything else you see that needs to be resolved before merging this?

davidelang:
On Fri, 15 Dec 2017, Rainer Gerhards wrote:

> rgerhards commented on this pull request.
>
>
>
>> @@ -605,8 +601,8 @@ Legacy Configuration Directives
>
>    Equivalent to: Ruleset
>
> -Legacy Example
> -^^^^^^^^^^^^^^
>
> I think it is too early now, but  in the end we should also remove obsolete examples. We need new style instead ;-)

We need to include basic examples, but can kill obsolete examples

rgerhards:
> We need to include basic examples, but can kill obsolete examples

I probably expressed myself unclear (and I guess this is what @davidelang  means):

* do **not** include any *obsolete* format examples
* we should **add** more *basic* and *advanced* format examples

rgerhards:
> What do you think about the state of the doc as of 0929512?

I think it is good to go. But for sure, we'll need to make changes in the future. Better merge early in any case.

Issue #441, opened 2017-12-17
Add "dynafile.donotsuspend" to global() config doc


deoren:
From rgerhards/rsyslog@e1758bf32cfc6eac9dd475324bc5a1e395aaf19d:

> omfile: add module-global option "dynafile.donotsuspend"
> 
> this permits to enable SUSPENDing dynafile actions. Traditionally,
> SUSPEND mode was never entered for dynafiles as it would have blocked
> overall processing flow. Default is not to suspend (and thus block).

PR #442, opened 2017-12-17
global: Add entry for "dynafile.donotsuspend" parameter

deoren:
Add details for new global parameter using commit log notes for the feature introduced on rsyslog/rsyslog#2236

closes rsyslog/rsyslog-doc#441


deoren:
@rgerhards Is this what you had in mind re coverage of the new parameter? I tried to use the same format that you used for the previous entry.

PR #417, opened 2017-12-10
alternative text for basic-advanced-obsolute format "name" (FOR REVIEW ONLY!)

rgerhards:
This does NOT use sphinx substitutions but instead uses the format
names also in the other text. The idea is to make the names sound
as natural and familar as possible.

PR #447, opened 2017-12-21
Hand merge existing work on PR 417

deoren:
@rgerhards I wasn't sure if you saw my notes on that PR asking where I should push my changes (a new branch, your branch, ...) so I went ahead and opted to push to a new branch on my end and submit this PR for your review.

Going forward, what would be the most useful approach for me to use to illustrate suggested changes against one of your PRs? We've both used in-line code reviews, but I am under the impression that those are mainly for short bits of feedback and not counter proposals for content changes?

refs rsyslog/rsyslog-doc#417

### DIFF
--- a/source/configuration/conf_formats.rst
+++ b/source/configuration/conf_formats.rst
@@ -5,61 +5,90 @@ Rsyslog has evolved over several decades. For this reason it supports three
 different configuration formats ("languages"):
 
 -  |FmtBasicName| - previously known as the :doc:`sysklogd  <sysklogd_format>`
-   format, this configuration format is widely known, actively supported and
-   occasionally enhanced with new features. This format is recommended for simple
-   use cases where the entire statement cleanly fits on a single line. Note
-   that some few constructs are no longer supported because they are incompatible
-   with newer features. These are mentioned in the compatibility docs.
+   format, this is the format best used to express basic things, such as where
+   the statement fits on a single line. It stems back to the original
+   syslog.conf format, in use now for several decades.
 
--  |FmtAdvancedName| - previously known as "``RainerScript``",
-   this is the current, best and most precise format to use for non-trivial use
-   cases where more than one line is needed. This format is available in
-   rsyslog v6 and above.
+   The most common use case is matching on facility/severity and writing
+   matching messages to a log file.
 
--  |FmtObsoleteName| - previously known simply as the ``legacy``
-   format, this format was created in the early days where we expected that
-   rsyslog would extend syslklogd just mildly. Consequently, it was primarily
-   aimed at small additions to the original sysklogd format. In essence,
-   everything that needs to be written on a single line that starts with a
-   dollar sign is legacy format. This format is notoriously difficult to use
-   and should not be used for new configurations. Users of this format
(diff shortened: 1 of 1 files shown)


deoren:
@rgerhards Welcome. Please let me know if there is a better way to handle these cases going forward and I will do so.

rgerhards:
> @rgerhards Welcome. Please let me know if there is a better way to handle these cases going forward and I will do so.

@deoren given the (inherent) limitations of github, I think this is the best path forward. It's not a real co-editing environment. If the "PR path" becomes to clumpsy, we may want to check out some collaborative editing tool in order to get to a PR, but if we can avoid it, I think it would be best...

rgerhards:
@deoren
> we may want to check out some collaborative editing tool in order to get to a PR,

It just occurred to me: maybe we could (ab)use the github wiki to craft complex text changes for a PR?

deoren:
@rgerhards Yeah, I think either of those could work. We heavily use an internal wiki at work to record our notes and have used pages to compare different approaches to documenting a process. Thus far it has worked well for our environment. 

I've also worked a little with others in Google Docs with both of us live editing content. That was a surreal experience; co-editing a standard doc via Google Docs actually felt too easy. :P If that approach were used for this type of work we'd then have to turn around and pull the content back out and format into reStructuredText, so unless it was small snippets it might not be worth using an outside tool. Granted, if we use a wiki page we would still have that work, but at least it would already be here in the project and not outside of it.

I'm fine with using PRs to offer contrasts with our proposed edits though. It worked well enough for this work and I can see it working equally well for other tasks.

I guess the takeaway here is that you're up for using an alternate approach and so am I, so we have options.

Just to fill in the gap in my knowledge, do I have the technical permission to update a branch in your repo (`rgerhards/rsyslog-doc`)? In other words, could I have pushed changes to the `rgerhards:alttry-fmt-names` branch to illustrate and then after we finished, merged from that branch instead of a separate branch in my repo?

rgerhards:
> I guess the takeaway here is that you're up for using an alternate approach and so am I, so we have options.

yup! (Google Docs is fine with me as well, just so you know).


> Just to fill in the gap in my knowledge, do I have the technical permission to update a branch in your repo (rgerhards/rsyslog-doc)? In other words, could I have pushed changes to the rgerhards:alttry-fmt-names branch to illustrate and then after we finished, merged from that branch instead of a separate branch in my repo?

good point. You are a maintainer here. So if I do not untick the default PR setting, it permits maintainer edits (I keep it ticked). So, yes, for PRs you can push back to mine. Didn't think about it, that might be a good solution (let's try it out to be sure).

deoren:
> @rgerhards good point. You are a maintainer here. So if I do not untick the default PR setting, it permits maintainer edits (I keep it ticked). So, yes, for PRs you can push back to mine. Didn't think about it, that might be a good solution (let's try it out to be sure).

I gave it a try on rsyslog/rsyslog#2283 to see how well that approach works. Any awkwardness to the approach was more me getting used to the commands needed to have git setup the branch so I could make the changes and commit back. I think I have it figured out.

PR #449, opened 2017-12-21
imklog: directive $klogLocalIPIF does not exist for several years now

rgerhards:


deoren:
@rgerhards

I looked through the [source](https://github.com/rsyslog/rsyslog/blob/2af5c24d2df128109855e21e4821ebab7b88f4b8/runtime/glbl.c#L1335) and found this:

```c
CHKiRet(regCfSysLineHdlr((uchar *)"localhostipif", 0, eCmdHdlrGetWord, setLocalHostIPIF, NULL, NULL));
```

Is `localhostipif` the advanced format equivalent of [the `$LocalHostIPIF` parameter](http://www.rsyslog.com/doc/v5-stable/configuration/global/index.html)?

It appears to be, but I didn't want to assume. If so, I don't see it listed on the `source/rainerscript/global` page.

source/configuration/modules/imklog.rst
```diff
@@ -35,12 +35,13 @@ Note: parameter names are case-insensitive.
...
 
-   If provided, the IP of the specified interface (e.g. "eth0") shall be
-   used as fromhost-ip for imklog-originating messages. If this
-   directive is not given OR the interface cannot be found (or has no IP
-   address), the default of "127.0.0.1" is used.
+   This directive is no longer supported. Instead, use the global
+
+   $localHostIPIF
+
+   directive instead.
```
deoren: What terminology do you prefer to be used, "parameter" or "directive"? Does it differ depending on the format? We discussed this previously on https://github.com/rsyslog/rsyslog-doc/issues/391#issuecomment-350490757, but we mixed in multiple topics on that ticket.

This is actually another good bullet list item for a CONTRIBUTING doc. On that note, I've started work on a _very_ early draft of such a document.

davidelang:
On Fri, 22 Dec 2017, Deoren Moor wrote:

> What terminology do you prefer to be used, "parameter" or "directive"? Does it differ depending on the format? We discussed this previously on https://github.com/rsyslog/rsyslog-doc/issues/391#issuecomment-350490757, but we mixed in multiple topics on that ticket.

directive made sense in the now obsolete config format where you set many things 
individually that then applied to a later config line

with the advanced config format, they would be parameters.

deoren:
> @davidelang: directive made sense in the now obsolete config format where you set many things
individually that then applied to a later config line
> 
> with the advanced config format, they would be parameters.

Thanks for the feedback. Oh well, sounds like the bulk of my changes in 5aab9129c5cccdb709183e5a753796d70fc5e423 were a bit misguided then.

PR #450, opened 2017-12-22
Add RainerScript is_time() function

mrworkman:
Adds documentation for the `is_time()` function.

See rsyslog/rsyslog#2246 and rsyslog/rsyslog#1624.

### DIFF
--- a/source/rainerscript/functions.rst
+++ b/source/rainerscript/functions.rst
@@ -363,3 +363,58 @@ produces
 ::
 
    1507151411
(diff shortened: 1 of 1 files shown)


deoren:
@mrworkman Looks good to me (for what that is worth), but I've asked @rgerhards to take a look to make sure I've not overlooked anything (e.g., `format_time` is marked as experimental, but not `parse_time` or `is_time`).

Issue #455, opened 2017-12-22
Prune copyright notices from individual pages


deoren:
From a discussion on #436:

> @rgerhards: I would also prune any remaining copyright notices on individual pages. They stem back to the original hand-crafted html doc and are no longer needed as sphinx generates the footer. Just nobody removed them so far, but it is fine to do so.

While this can certainly be done as a "when you encounter it" item, it may be worth a focused mass prune effort.

Issue #452, opened 2017-12-22
Update hard-coded html page references to use Sphinx's doc cross-referencing syntax instead


deoren:
Example change:

```diff
-see our** `troubleshooting guide <troubleshooting/index.html>`_ **to get started.**
+see our** :doc:`troubleshooting guide <troubleshooting/index>` **to get started.**
```

rgerhards:
all of those things are unfortunate left-overs from the hand-crafted HTML doc (just to note the obvious).

deoren:
Output from running epubcheck (current master branch of IDPF/epubcheck repo) against 8ff5cdf068daa674a59e9b5d4c901b616d9dc3f3:

`deoren@ubuntu-testbox:~/dev/rsyslog/rsyslog-doc$ java -jar ../../epubcheck/target/epubcheck.jar build/rsyslog.epub`

```
Validating using EPUB version 3.0.1 rules.
ERROR(CSS-008): build/rsyslog.epub/_static/rsyslog.css(52,12): An error occurred while parsing the CSS: Token '{' not allowed here.
ERROR(CSS-008): build/rsyslog.epub/_static/rsyslog.css(52,41): An error occurred while parsing the CSS: Token '}' not allowed here.
ERROR(RSC-007): build/rsyslog.epub/community.xhtml(46,77): Referenced resource 'licensing.html' could not be found in the EPUB.
ERROR(RSC-007): build/rsyslog.epub/community.xhtml(48,85): Referenced resource 'free_support.html' could not be found in the EPUB.
ERROR(RSC-007): build/rsyslog.epub/features.xhtml(24,125): Referenced resource 'rsyslog_ng_comparison.html' could not be found in the EPUB.
ERROR(RSC-007): build/rsyslog.epub/features.xhtml(30,83): Referenced resource 'rsyslog_mysql.html' could not be found in the EPUB.
ERROR(RSC-007): build/rsyslog.epub/features.xhtml(35,76): Referenced resource 'ommail.html' could not be found in the EPUB.
ERROR(RSC-007): build/rsyslog.epub/features.xhtml(40,113): Referenced resource 'rsyslog_high_database_rate.html' could not be found in the EPUB.
ERROR(RSC-007): build/rsyslog.epub/features.xhtml(63,74): Referenced resource 'rsyslog_tls.html' could not be found in the EPUB.
[...]

PR #458, opened 2017-12-24
pmlastmsg: Fix links, prune copyright

deoren:
refs rsyslog/rsyslog-doc#452, rsyslog/rsyslog-doc#455


deoren:
@rgerhards Just to make sure I understand, but the goal for #455 is to nuke both the copyright notice _and_ the link back to the docs (relying on an auto-generation of that link), right?

rgerhards:
> @rgerhards Just to make sure I understand, but the goal for #455 is to nuke both the copyright notice and the link back to the docs (relying on an auto-generation of that link), right

yup. I just didn't think of the backlink, but that has to go away as well!

PR #468, opened 2018-01-03
re-add release_build.sh script: used by maintainer during release

rgerhards:
the build.sh does too many things that are not directly related
to release build process
  
closes https://github.com/rsyslog/rsyslog-doc/issues/469

### DIFF
--- a/release_build.sh
+++ b/release_build.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+# this is used to create an rsyslog release tarball.
+# it must be executed from the root of the rsyslog-doc
+# project.
+
+docfile=rsyslog-doc.tar.gz
+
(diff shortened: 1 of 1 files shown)


rgerhards:
@deoren let me know if you have concerns adding this back. If so, we should change ```build.sh``` to just do the simple steps that this script did. Alternatively, I could keep it in my personal archive, but that's a bit clumsy (albeit might be the right thing todo except for me lazyness... ;-)

deoren:
@rgerhards No concerns about bringing it back. It took me a while to find the commit (longer than it should have), but it looks like the file last existed in 3ebd747e07ac731f3b2449d298facb2843517f14.

I restored it to my branch, but will skip it since you've already done so here. I'll also rename the build.sh script.

Issue #469, opened 2018-01-03
build script: restore old build script used for release builds


deoren:
refs  #424, #464, #465

PR #471, opened 2018-01-03
Rename build script to better reflect purpose

deoren:
The prior name gave the wrong impression of its purpose. This new name is an attempt to clarify the purpose.

refs rsyslog/rsyslog-doc#469


deoren:
@rgerhards Thanks.

PR #465, opened 2017-12-31
build.sh: Fix typos, block accidental commits

deoren:
@rgerhards I found some issues with the build script when attempting to use it as-is for automating regular builds of the branches. I'm not sure that I have all of them squashed, but this fixes at least the most glaring ones. I suspect I'll loop back to this script and make some further tweaks as part of another PR after I have a chance to spend some more time with it.

closes rsyslog/rsyslog-doc#464

Issue #424, opened 2017-12-11
Update docs build process to dynamically version output based off of info in Git


deoren:
Instead of having to manually update the `sources/conf.py` file or the `build.sh` script to reflect the version number/branch, have both files dynamically generate needed values.

deoren:
Sorry for the squash/push noise. I didn't realize that would happen when referring to an open ticket via a commit message. I assumed that the ticket updates would be similar to the PR updates where only the unsquashed commits would show.

PR #472, opened 2018-01-03
Restore build.sh from 3ebd747

deoren:
Restore the prior stable version of this script since the most recent copy of this script was recently labeled specifically as a cron script

refs rsyslog/rsyslog-doc#468, #469, #465, #424

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+
+echo "This script is intended to run with the clean repo version of the code."
+echo "Run the sphinx-build command manually if you want to see your uncommited changes."
+echo "If you run this command with uncommitted and un-pushed changes, you will lose those changes,"
+echo "or end up with merge conflicts."
+echo ""
+echo "Press Enter to continue or Ctrl-C to cancel...."
+read
+
+MASTERBRANCH=v8-devel
+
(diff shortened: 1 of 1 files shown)


deoren:
@rgerhards If you don't restore it, perhaps modify the `tar` command in your release script to exclude it?

rgerhards:
Ahh!!! I didn't even notice that. Some thing for the renamed one: should we include it? Do we consider it "part of the doc" or more "ops utility primarily for the rsyslog-doc team"?

deoren:
> @rgerhards: Some thing for the renamed one: should we include it? 

I don't think so. It remains in the repo for anyone that wishes to use it and the README file covers the steps to build from master if someone wishes to do that themselves. Since it is included in the tarball that aspect should be covered.

> @rgerhards: Do we consider it "part of the doc" or more "ops utility primarily for the rsyslog-doc team"?

I think the latter.

At some point we may wish to create a `contrib` directory and place files like these build scripts there along with a small README noting their purpose. I have a `/etc/cron.d/rsyslog-doc` and `/use/local/sbin/rsyslog-doc-build.sh` script that could be added there also.

deoren:
I'll go ahead and close this out and will include the change to the `release_build.sh` script as part of a different PR.

PR #476, opened 2018-01-05
PROTOTYPE: Parameter formatting

deoren:
This is early work to prototype a consistent format for module parameters. 

Some of the formats I've observed:

- Bullet points
- use of  `.. function::` directives with "default" statement just beneath
- ... with "default" statement at the end
- bullet point, noting whether parameter is mandatory and then default value (one-liner)

I don't know what the right answer is, so this PR is more of a fence post for conversation than anything else. Because I recently worked on the imuxsock module doc, I'll test some changes to that doc first.

- Refs: #6

---

Live view of work:

- [section headers with `obsolete legacy` format directives](http://rsyslog.whyaskwhy.org/imuxsock_parameter_syntax/configuration/modules/imuxsock.html)

PR #479, opened 2018-01-08
Attempt to clarify SysSock.Use parameter use

deoren:
> Instead of trying to give too many details in the parameter entry, forward the reader to the dedicated section for the topic of systemd and imuxsock interaction.

@rgerhards I kept getting hung up on the existing wording while working on variations of the prototype formats in PR #476. I'm hoping this simpler approach will read better and let the dedicated module doc section handle stepping the reader through the interaction with systemd (vs trying to replicate it in multiple places).

### DIFF
--- a/source/configuration/modules/imuxsock.rst
+++ b/source/configuration/modules/imuxsock.rst
@@ -60,13 +60,10 @@ Global Parameters
    ``SysSock.Unlink`` setting, this socket is created upon rsyslog startup
    and deleted upon shutdown, according to traditional syslogd behavior.
 
-   The behavior of this parameter is different for systemd systems. For those
-   systems, ``SysSock.Use`` still needs to be enabled, but the value of
-   ``SysSock.Name`` is ignored and the socket provided by systemd is used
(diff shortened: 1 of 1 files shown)


rgerhards:
I like it!

rgerhards:
@deoren Not sure if you considered it work in progress. I'd say lets merge frequently, so I would merge if you don't object.

deoren:
> @rgerhards: I like it!

thanks

> @rgerhards: @deoren Not sure if you considered it work in progress. I'd say lets merge frequently, so I would merge if you don't object.

I was considering it a small one-off change, but didn't want to merge until you had a chance to look at it. Going forward, I'll go ahead and merge small changes as we talked about and will just back them out if you or others object to them. Also as discussed (just being clear), I'll seek input for larger changes before merging them myself (#476 is a recent example).

In general though, I like to make frequent small changes as they're easier to back out and (probably more importantly) it gives the team a chance to offer feedback before a larger batch of modifications are attempted at once.

Issue #482, opened 2018-01-08
release process: need way to specify version number


rgerhards:
The improvements to the doc generation now generate a hash-based version number. That is great for devel releases, but we need to find a way to use the regular version number (e.g. "8.32.0") for release builds.

deoren:
@rgerhards

From the first [build script](https://github.com/deoren/rsyslog-examples/blob/master/scripts/build_docs.sh) that I used:

```bash
version=$(grep -E '^version' source/conf.py | awk '{print $NF}' | sed "s/'//g")
datestamp=$(date '+%Y%m%d')
commit=$(git describe | awk -F'-' '{print $NF}')
release="${version}.${datestamp}.${commit}"

declare -a formats
formats=(
  epub
  html
)

for format in "${formats[@]}"
do
    sphinx-build -D version="$version" -D release="$release" -b $format source build
done
```

I eventually worked that into the `build.sh` script with the idea that someone could run that script whenever needed to generate fresh docs from all branches. It ended up growing in a different direction than originally intended when I made the mistaken assumption that the `release_build.sh` script could be supplanted by a modified `build.sh` script that did everything both of the original scripts did.

An earlier form of the combined script actually did work well enough for that, but then I started work to get the script into shape for automation. My thoughts at the time were that the `build.sh` script could be used on a build server to periodically generate fresh copies of all branches *and* generate a tarball that could be used for official stable releases.

I *meant* to follow up with you about setting up such an automated build process, but again I made the mistake of assuming that there was a different/official process for releasing the stable doc tarball outside of the repo. That proved to be false and on hindsight wasn't a good assumption (it makes sense to keep the scripts in the repo and use an include mechanism for anything sensitive).

I'll submit a PR for a small tweak to the `release_build.sh` script that references a script argument for setting the version info.

On a related note, I've setup such an automated build process on one of my VPS boxes in an effort to help with sharing prototype changes as I work on PRs. I've not yet provided a PR for the `/etc/cron.d/rsyslog-doc` or `/usr/local/sbin/rsyslog-doc.sh` files, but plan to do so once I let the process run for a while longer. At that point setting up a `contrib/scripts` dir might be the way to go.

deoren:
@rgerhards I forgot to clarify in the last response: That small script example uses a hash-based version numbering for the builds, but shows via the `sphinx-build` calls how to go about using a stable version number via `sphinx-build` command-line options.

davidelang:
can we use git describe for this?

[...]

PR #483, opened 2018-01-08
Update build script to accept version/release info from command-line

deoren:
Add support for pulling in version number from command-line. This can be used in overriding the hash-based version number now used by default for devel builds.

Additionally, the release build script no longer attempts to include build scripts in the doc release tarball.

closes rsyslog/rsyslog-doc#482

### DIFF
--- a/release_build.sh
+++ b/release_build.sh
@@ -31,7 +31,7 @@ sphinx-build -D version="$version" -D release="$release" -b $format source build
 
 [ ! -e ./$docfile ] || rm -rf ./$docfile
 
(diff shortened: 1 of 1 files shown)


rgerhards:
thx for the very quick action!

BTW: I do not do an explicit "go, all OK" review for those simple things. But I review before I merge ;-)

rgerhards:
That would be fine with me. (On phone)

Sent from phone, thus brief.

Am 09.01.2018 13:56 schrieb "Deoren Moor" <notifications@github.com>:

PR #486, opened 2018-01-09
release_build: create tarball with versioned name

rgerhards:
now that we have the version argument, we can directly
create the right tarball name ;-)

### DIFF
--- a/release_build.sh
+++ b/release_build.sh
@@ -15,7 +15,7 @@ fi
 
 release="${version}.0"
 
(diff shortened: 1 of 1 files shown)


deoren:
I thought that the doc packaging process required a static tarball name (without a version number)?

rgerhards:
No, i renamed it manually ;-)

deoren:
Oh. Sorry, had I realized that I would have handled that in my PR as well. Now I know. :)

Issue #491, opened 2018-01-11
create documentation on rsyslog use in containers


rgerhards:

deoren:
Since this is an long-running ticket on the same topic, I'll post the suggestions I made as doc comments here.

These remarks are in regards to the `source/containers/container_features.rst` file.

> Upon start, rsyslog checks if it is running as pid 1. If so, this is a clear indication it is running inside a container.

- Explain why this is a clear indication. Perhaps make a quick reference to what normally holds pid 1 for anyone new to UNIX process management

> This changes some parameter defaults

- Add details explaining WHY the parameter defaults were changed.

refs #494

deoren:
TODO: Reference official Docker Hub (page, profile, id?): 

https://hub.docker.com/u/rsyslog/

rgerhards:
I'll add it when I finally modify the PR ;-)

deoren:
> @rgerhards: I'll add it when I finally modify the PR ;-)

Yep. Is it helpful to add standalone/misc items like that to existing tickets (which appear to be works in progress) or as separate tickets? I was thinking the former based on the state.

Issue #451, opened 2017-12-22
Create doc to outline basic guidelines for the documentation


deoren:
The intention is to provide a set of guidelines to be used when creating new documentation or refreshing older content.

Example of items that would be useful to cover:

- Older content that should be removed when encountered
- Substitution references
    - where are they defined?
    - why and where should they be used?
- Standard section headers
- Terminology
    - "parameter" vs "directive" ?
- Include files
    - which ones are available?
    - where they should be used?

rgerhards:
I find this very useful, thx for the initiative.

> "parameter" vs "directive" ?

For the obsolete format, I often spoke (and still tend to speak) of "directives" because (to me ;-)) they have something command-like.

For new style, we have configuration **objects** and objects have **parameters** (ok, properties, but IMHO parameters is easier to grasp for a config language).

I have no objection if we call everything (obsolete and new) "parameter". I just need to get used to it and it helps if we write it down. Writing down is key in any case, as the currently existing inconsistency is from not knowing 100% sure which term to use.

deoren:
> @rgerhards I have no objection if we call everything (obsolete and new) "parameter". I just need to get used to it and it helps if we write it down. 

@davidelang had some good feedback on https://github.com/rsyslog/rsyslog-doc/pull/449#issuecomment-353711372:

> directive made sense in the now obsolete config format where you set many things
> individually that then applied to a later config line
> 
> with the advanced config format, they would be parameters.

Perhaps my efforts to unify the two via 5aab9129c5cccdb709183e5a753796d70fc5e423  was a bit misguided. Since there is ample history of referring to the obsolete legacy format "settings" as directives, it may be a good idea to keep that term associated with them. As we move further away from the format and purge the bulk of the references from the current documentation it would be easier to spot inconsistencies in the remaining docs.

> @rgerhards Writing down is key in any case, as the currently existing inconsistency is from not knowing 100% sure which term to use.

I agree.

deoren:
#504 brought up some good points for clarification:

## Should the docs tarball provide sources within the tarball or assume users will pull from the Git repo? 

- Precedence and the license likely require that sources be included

- The general expectation is that sources will build without a dependency on a remote repo. We should note this requirement in the doc. This should have probably been obvious, but it's best to be explicit to help prevent repeat issues for downstream users/package maintainers.

## What should the behavior be for obtaining the doc tarball, decompressing it and then running `sphinx-build -b html source build`?

[...]

Issue #438, opened 2017-12-16
Update header levels to be consistent


deoren:
While working on some module docs recently I noted that the header level syntax was inconsistent between at least two of them. If two, then likely more.

## Task list

- [ ] Define section headers (style guide?)
- [ ] Create template for new pages (perhaps a template per page type?)

## Guides for section headers

1. http://www.sphinx-doc.org/en/stable/config.html#confval-text_sectionchars
1. http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#sections
1. https://devguide.python.org/documenting/#sections

## Notes

The docutils page confirms what characters are valid for forming section headers:

> The following are all valid section title adornment characters:
> 
> ``` ! " # $ % & ' ( ) * + , - . / : ; < = > ? @ [ \ ] ^ _ ` { | } ~ ```
> Some characters are more suitable than others. The following are recommended:
> 
> ``` = - ` : . ' " ~ ^ _ * + # ```

The Python guide (link 3) is a little more specific with their suggested convention:

> `#` with overline, for parts
> `*` with overline, for chapters
> `=`, for sections
> `-`, for subsections
> `^`, for subsubsections
> `"`, for paragraphs

The `text_sectionchars` option on the Sphinx doc page is intended for text output:

> A string of 7 characters that should be used for underlining sections. The first character is used for first-level headings, the second for second-level headings and so on.
> 
> The default:
> 
>   ```  '*=-~"+`'.  ```

Even so, it provides a starting point as a standard for setting section header levels.

deoren:
After giving this some more thought, I think it may be worth sticking with the established style that the Sphinx project uses for their own documentation. I'm not a fan of the characters chosen for the levels, but if we can simply reference an existing style used by multiple large projects (which is already in their docs as a suggested approach), it should make it easier to work with new contributors (existing reference material to point them to).

rgerhards:
Actually, I admit that the style currently used in rsyslog is not consistent. I myself would appreciate if we had a simple README or CONTRIB  (or whatever) document that spells out how the rsyslog-doc project uses it. TBH, I do not really care what it is, as long as I know where I can find what to use. Of course, it does help if we follow some established and well-known policy ;-) Just my personal 2cts (NOT speaking as rsyslog maintainer here!).

deoren:
> @rgerhards: I myself would appreciate if we had a simple README or CONTRIB (or whatever) document that spells out how the rsyslog-doc project uses it. 

[...]

PR #492, opened 2018-01-11
begin documentation on rsyslog containers

rgerhards:
The idea is to provide useful information as early as possible and
extend it as it grows. While this at first may look sparse, I think
it is the best way forward. The alternative would be to wait a
long time until we know "everything" and can document it.

see also https://github.com/rsyslog/rsyslog-doc/issues/491

### DIFF
--- a/source/containers/container_features.rst
+++ b/source/containers/container_features.rst
@@ -1,0 +1,9 @@
+Container-Features
+------------------
+
+Upon start, rsyslog checks if it is running as pid 1. If so, this is
+a clear indication it is running inside a container. This changes
+some parameter defaults:
+
+- ctl-c is enabled
+- no pid file is written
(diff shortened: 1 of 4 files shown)


rgerhards:
@deoren I have begun to document some findings and important new 8.33 features for containers. Not sure about the exact placement in the TOC, and what the recommended writing style (header levels etc) this is. Would appreciate suggestions.

I think it makes sense to merge early, as this is information people are looking after - especially the potential problem part.

deoren:
@rgerhards Awesome. Since we've yet to work out the details re writing style, I think it's fine to merge as-is. I'm hoping that the work on #476 will help work out some of those specifics. I can then take what I've learned there and can work on closing out #438 and then move forward with work on #451.

Regarding placement, let me get used to seeing the current placement in the master branch as I work on other things and I can better give feedback then. Like you, I also think merging the work in early makes sense. Others that are looking for it can quickly benefit from what you've discovered and can offer suggestions for improvement as well.

PR #494, opened 2018-01-12
Minor tweaks to Container-Features doc

deoren:
- Typo fix
- Use 'kbd' Sphinx directive to highlight the specific keystroke used to exit rsyslog
- Add comments noting points for future expansion

### DIFF
--- a/source/containers/container_features.rst
+++ b/source/containers/container_features.rst
@@ -5,5 +5,5 @@ Upon start, rsyslog checks if it is running as pid 1. If so, this is
 a clear indication it is running inside a container. This changes
 some parameter defaults:
 
-- ctl-c is enabled
+- :kbd:`Control-C` is enabled
 - no pid file is written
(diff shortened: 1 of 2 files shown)


deoren:
@rgerhards What are your thoughts re TODO comments (those that do not render where the user sees them)? In this case I was looking to add markers for further expansion. Do you feel those remarks belong in the docs or should they only be handled as new issues here?

rgerhards:
I would go for using issues. A strong motivation for this is:

* we can assign milestones
* we can use github project tracking
* **this is more visible/appealing to potential (new) contributors IMHO**

deoren:
> @rgerhards: I would go for using issues. A strong motivation for this is:
> 
> - we can assign milestones
> - we can use github project tracking
> - this is more visible/appealing to potential (new) contributors IMHO

Good points. I'll extract the comments and create separate issues. Once the comments are out I'll squash commits and repush the changes to the associated branch.

deoren:
Updated commit to exclude doc comments.

PR #497, opened 2018-01-12
Create requirements.txt, update travis conf

deoren:
With some of the features I am working with we may need to get specific with the package requirments. This is a prototype to make sure that I have the syntax right before merging to master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: python
 python: "2.7"
 
 install:
-  - pip install -q sphinx
+  - pip install -r requirements.txt
 
 script:
   - sphinx-build -n -W -b html source build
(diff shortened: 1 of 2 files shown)


deoren:
@rgerhards Any concerns with this change?

rgerhards:
@deoren I just keep the PR open to make sure you actually notice my comment. Feel free to merge yourself, I am fine with that.

deoren:
> @rgerhards: Maybe it would be a good amendment to README how to install that version on popular distros for the non-pythonish? I guess install of pip followed by pip install should generally do it on all distros...

Good idea. I'll get that in. I'll create a separate PR for that and will merge it before this one (so I don't get distracted and forget).

Issue #500, opened 2018-01-12
impstats module doc missing example output, explanation for the 'size' counter


deoren:
AFAIK, the only place that mentions that `size` is an integer listing the number of messages passing through the associated object is this page:

http://www.rsyslog.com/periodic-statistics-on-rsyslog-counters/

The doc could use other general cleanup as well, but adding an example of the output and an explanation that of the `size` counter would be of more immediate use.

PR #501, opened 2018-01-12
impstats: Headers, trim legacy, example output

deoren:
- Swap out existing section header characters for those used in Python Style Guide and the Sphinx documentation
- Add example output from module
- Add explanation of 'size' counter, emphasize that it tracks log messages flowing through the associated object
- minor formatting tweaks (still much to do)
- Purged obsolete legacy configuration examples, sections
  since there are advanced format examples already included

refs rsyslog/rsyslog-doc#438
refs http://www.rsyslog.com/periodic-statistics-on-rsyslog-counters/
closes rsyslog/rsyslog-doc#500

### DIFF
--- a/source/configuration/modules/impstats.rst
+++ b/source/configuration/modules/impstats.rst
@@ -1,5 +1,5 @@
 impstats: Generate Periodic Statistics of Internal Counters
-===========================================================
+***********************************************************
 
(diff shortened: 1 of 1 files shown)


deoren:
Local build passes checks, so going ahead and merging this. There is further cleanup work that can be done here, but I don't have the time to invest in it at the moment. We can apply further fixes in a future PR.

Issue #504, opened 2018-01-14
building from source taball is broken


mbiebl:
Since commit de28b1a5d2f9c5b606ca4065354f1578f7e9cf61 building the documentation from the source tarball is broken, as the source tarball does not have a .git directory:
```
Running Sphinx v1.6.6
fatal: Not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

Configuration error:
There is a programable error in your configuration file:

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/sphinx/config.py", line 157, in __init__
    execfile_(filename, config)
  File "/usr/lib/python3/dist-packages/sphinx/util/pycompat.py", line 150, in execfile_
    exec_(code, _globals)
  File "conf.py", line 69, in <module>
    git_tag_output = subprocess.check_output(['git', 'tag', '--list', "v*"]).decode("utf-8").strip()
  File "/usr/lib/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['git', 'tag', '--list', 'v*']' returned non-zero exit status 128.
```

mbiebl:
cc @deoren

mbiebl:
btw, it's not a good idea to use the current day in the release version like this

release = "{}-{}-{}".format(version, TODAY, commit)

The release version should not changed depending on whether I built the docs yesterday or today.

mbiebl:
See https://reproducible-builds.org/docs/timestamps/ why embedding timestamps is a bad idea

deoren:
@mbiebl 

Thanks for the report and sorry for the trouble you experienced.

I'm actively researching and drafting a detailed response; I will respond back once I've got the details together.

mbiebl:
fwiw, I applied this patch for now in Debian: https://salsa.debian.org/debian/rsyslog-doc/commit/fdb2275fdcf1252af969cc2a2908ce019595686f

Manually updating the version number is obviously not a proper fix, so I hope we can find a solution which works for everyone.

deoren:
@mbiebl What is your build process for producing the release tarball?

mbiebl:
[...]

PR #493, opened 2018-01-11
Update build process to default to static value builds, use verbose details for dev builds when available

deoren:
This is intended for use with dev builds of the docs to quickly identify what branch the docs were bult from. This also helps with epub builds as some readers do not always display the "release" details prominently once the file has been opened.

Note: Looks like my editor applied some whitespace trimming when I saved the index file, so the commit isn't entirely "clean".

@rgerhards What do you think? The results should be visible here soon (builds are performed every 15 minutes, takes just under 10 minutes to complete):

http://rsyslog.whyaskwhy.org/display_branch_build_info/

PR #505, opened 2018-01-14
Sphinx build conf: backout de28b1a changes

deoren:
Commit de28b1a introduced dynamic versioning of content, but failed to take into account the scenario of building the docs from the tarball instead of from a local git repo.

As a quick fix, this commit changes the conf back to an equivalent static version so that the doc sources build as-is from the doc tarball.

I plan to restore dynamic builds in a future PR (likely #493), but this should solve the immediate problem.

refs rsyslog/rsyslog-doc#504, rsyslog/rsyslog-doc#424


deoren:
@rgerhards Because I'm replacing (at least temporarily) what has proven to be an unstable change to something stable/simple, I'm going to ahead and merge this. 

Please let me know whether this was a problem. I'm working to get PR #493 into a ready state for inclusion soon and I think that will solve some of the aggravation around what I (ironically) intended to be an enhancement to the build process.

I'll rebase, push and remove the blocker label on that PR once the related changeset is ready for review.

rgerhards:
> Please let me know whether this was a problem.

I don't think it was a problem, but I also think it wasn't that urgent. The release tarball is built on release, and the next one is 5 weeks away. At least I would not consider this an emergency to do a re-release. 

@mbiebl do you consider it worth a re-release of the tarball?

mbiebl:
No, as I said, it's not urgent from my side. I already worked around that in the Debian package. 
I just thought I should file that while packaging the latest upstream release.
Please take your time for fixing this properly. There is no need for a new release tarball or urgent changes.

rgerhards:
@mbiebl thanks - your continuous help is much appreciated!

deoren:
@rgerhards @mbiebl Thank you both for your feedback. It sounds like we're good with completely resolving this with the next release then.

PR #507, opened 2018-01-18
Explain use of key to get random-consistent partition choice

matthiasr:
It took a while to figure out how to do this, so document the findings
for everyone.

### DIFF
--- a/source/configuration/modules/omkafka.rst
+++ b/source/configuration/modules/omkafka.rst
@@ -73,6 +73,9 @@ comma-delimited list of values as shown here:
 
    Kafka key to be used for all messages.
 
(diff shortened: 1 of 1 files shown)


matthiasr:
One caveat is that I don't know when this started working. I noticed on a node that had rsyslog 8.31 and an unknown version of librdkafka this configuration caused the output to be suspended, but an upgrade made it work.

deoren:
@matthiasr Thank you for the PR!

Did you figure out the logic by trial and error, a review of the source or another method? Just curious whether we should note any reference works used, perhaps as a "seealso" entry for further reading.

deoren:
@matthiasr Not a problem, just wanted to double-check. Thank you for sharing your findings back as doc improvements!

PR #509, opened 2018-01-21
omrelp: Add RebindInterval parameter

deoren:
Copied from omfwd module doc and slightly modified to remove mention of TCP/UDP specifics.

My assumption (which is why I'm not going ahead and merging this), is that the behavior between the RebindInterval parameter for omrelp and omfwd is the same, but I don't know that for sure.

I plan to rework the formatting for this doc once PR #476 is reviewed/merged, but for now I just used the existing format in order to go ahead and solve the reported issue.

closes rsyslog/rsyslog-doc#345

### DIFF
--- a/source/configuration/modules/omrelp.rst
+++ b/source/configuration/modules/omrelp.rst
@@ -33,6 +33,14 @@ Note: parameter names are case-insensitive.
    considered dead and tried to recover.
 -  **conn.timeout** (not mandatory, default 10)
    Timeout for the socket connection.
+-  **rebindInterval** integer
+   Permits to specify an interval at which the current connection is
+   broken and re-established. This setting is primarily an aid to load
+   balancers. After the configured number of messages has been
+   transmitted, the current connection is terminated and a new one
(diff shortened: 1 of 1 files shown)


deoren:
@rgerhards Was I correct re the behavior of this parameter being the same between the omrelp and omfwd modules?

rgerhards:
> @rgerhards Was I correct re the behavior of this parameter being the same between the omrelp and omfwd modules?

that yes, but I need to check the changelogs if the bug in omrelp is fixed. think so, but better be sure...

deoren:
> @rgerhards: that yes, but I need to check the changelogs if the bug in omrelp is fixed. think so, but better be sure...

Alright. If you see anything else that needs changing just let me know.

Thanks.

rgerhards:
all is well, bug in omrelp rebindinterval was fixed in 8.30.0. So this PR should be good to go. Sorry for the long delay...

rgerhards:
well... no reason not to merge, so I do it now. Let's see some progress :-)

Issue #460, opened 2017-12-25
"random()" missing from functions list


deoren:
From https://github.com/rsyslog/rsyslog/issues/2224#issuecomment-353779101:

> @rgerhards: actually, random() exists, I think for quite a while. Maybe doc is missing (could not check right now).

Looking at the docs, I don't see it listed:

- http://www.rsyslog.com/doc/v8-stable/rainerscript/functions.html
- http://www.rsyslog.com/doc/master/rainerscript/functions.html

deoren:
If I'm reading the commit logs correctly, it appears that the `random` function was added in rsyslog/rsyslog@4db25fb98bf8f69ad847ae19b5e30aedee31f6fa

deoren:
NVM, I found mention of it even before then in rsyslog/rsyslog@59b46975bde1ac21730ae678935c3537babfbac8

deoren:
```c

/* grammar/rainerscript.c */

static long long
doRandomGen(struct svar *__restrict__ const sourceVal) {
	int success = 0;
	long long max = var2Number(sourceVal, &success);
	if (! success) {
		DBGPRINTF("rainerscript: random(max) didn't get a valid 'max' limit, defaulting random-number "
			"value to 0");
		return 0;
	}
	if(max == 0) {
		DBGPRINTF("rainerscript: random(max) invalid, 'max' is zero, , defaulting random-number value to 0");
		return 0;
	}
	long int x = randomNumber();
	if (max > MAX_RANDOM_NUMBER) {
		DBGPRINTF("rainerscript: desired random-number range [0 - %lld] "
			"is wider than supported limit of [0 - %d)\n",
			max, MAX_RANDOM_NUMBER);
	}
	return x % max;
}
```

@rgerhards I see how the function handles receiving no arguments and 0 as the argument (0 is returned), but I can't understand what is happening when a value is provided larger than the platform supports. I see that the `DBGPRINTF` statement is triggered to log the warning in the debug output, but I don't see where any action is taken to handle the "value too large" situation.

Is this because using the modulus operator is able to safely handle values larger than RAND_MAX (which MAX_RANDOM_NUMBER "points" to)? Is the only concern just logging why `return x % max` is potentially returning unexpected results?

rgerhards:
[...]

PR #510, opened 2018-01-21
Add random function coverage

deoren:
Add coverage of `random()` function. Not yet complete, waiting on feedback for related ticket.

closes rsyslog/rsyslog-doc#460

### DIFF
--- a/source/rainerscript/functions.rst
+++ b/source/rainerscript/functions.rst
@@ -225,11 +225,17 @@ random(max)
    While the original intent of this function was for load balancing, it
    is generic enough to be used for other purposes as well.
 
+.. warning::
+   The random number must not be assumed to be crypto-grade. rsyslog will
+   attempt to use a crypto-grade source if available, but it will fall
+   back to a very weak random number generator if a crypto-grade source
+   is not available.
+
 .. versionadded:: 8.12.0
 
 .. versionchanged:: 8.31.0
-   ``/dev/urandom`` is used if available. If it is not, an error is logged
-   and the weak PRNG is used instead.
+   If a crypto-grade random number generator is not available, rsyslog will
+   fall back to using a weak PNRG.
 
 ltrim
 -----
--- a/source/rainerscript/functions.rst
+++ b/source/rainerscript/functions.rst
@@ -210,6 +210,27 @@ ipv42num
    contain any other characters in between (except dots). If the format does include these, the
    function results in an error and returns -1.
 
+random(max)
+-----------
+
+   Generates a random number between 0 and the number specified, though
+   the maximum value supported is platform specific.
+
+   - If a number is not specified then 0 is returned.
+   - If 0 is provided as the maximum value, then 0 is returned.
+   - If the specified value is greater than the maximum supported
+     for the current platform, then rsyslog will log this in
+     the debug output and use the maximum value supported instead.
+
+   While the original intent of this function was for load balancing, it
+   is generic enough to be used for other purposes as well.
+
+.. versionadded:: 8.12.0
+
+.. versionchanged:: 8.31.0
+   ``/dev/urandom`` is used if available. If it is not, an error is logged
+   and the weak PRNG is used instead.
+
 ltrim
 -----
 


deoren:
@rgerhards What do you think of the current state of the PR? I backed out a commit where some unrelated whitespace changes were made to make the changes clearer.

deoren:
Rebased/pushed to squash the multiple commits tied to this PR. Unless I have something off with the coverage, this is probably ready to merge.

@rgerhards Any concerns with the coverage as of 4bd9d96af8f347f8aae1b97293d794e444960340?

rgerhards:
small but important comment: I think we should stress that this is not a crypto-grade random number. It actually currently is in most cases, but we do not want to guarantee that. Something along the lines of

"Important note: the random number must not be assume to be of crypto-grade. It may be generated by a very weak random number generator."

Thinking about this, I think we should probably better drop the note on /dev/urandom, as this implies crypto-grade. @deoren what do you think? Am I going overboard? But week PRNG has been too often source of real problems if assumed to be secure....

deoren:
> @rgerhards: small but important comment: I think we should stress that this is not a crypto-grade random number. It actually currently is in most cases, but we do not want to guarantee that. Something along the lines of
> 
> "Important note: the random number must not be assume to be of crypto-grade. It may be generated by a very weak random number generator."
> 
> Thinking about this, I think we should probably better drop the note on /dev/urandom, as this implies crypto-grade. @deoren what do you think? Am I going overboard? But weak PRNG has been too often source of real problems if assumed to be secure....

**Background:**

The change for 8.31.0 specifically mentioned using crypto-grade encryption if it was available and then falling back to PRNG if not. That is where I pulled that statement. 

**Thoughts:**

I believe that mentioning the specifics int he changelog is good, but I see your point regarding giving too much detail in the user doc. The doc is after all intended to reflect the interface, not the implementation. I sometimes forget that when pursuing to cover a feature in detail. I think this is a case where too much detail is not helpful (not "bad", but not really useful either).

I've reworked the wording and added an explicit warning re suitability of the result for crypto use.

rgerhards:
> Background:
> 
> The change for 8.31.0 specifically mentioned using crypto-grade encryption if it was available and then falling back to PRNG if not. That is where I pulled that statement.

haha, looks like my own fail ;-)

> 
> Thoughts:
> 
> I believe that mentioning the specifics int he changelog is good, but I see your point regarding giving too much detail in the user doc. The doc is after all intended to reflect the interface, not the implementation. I sometimes forget that when pursuing to cover a feature in detail. I think this is a case where too much detail is not helpful (not "bad", but not really useful either).

Sometimes it is even bad -- because if we document it, we must support it in the future. For example, a PRNG crypto-grade gurantee would be a real problem on e.g. Solaris, where it is hard to obtain. Here we fall back to just the clib, which is very far from crypto grade. It's no problem, because this is not defined to be crypto grade.

> 
> I've reworked the wording and added an explicit warning re suitability of the result for crypto use.

Looks much better to me :-)

deoren:
> @rgerhards: Sometimes it is even bad -- because if we document it, we must support it in the future. For example, a PRNG crypto-grade gurantee would be a real problem on e.g. Solaris, where it is hard to obtain. Here we fall back to just the clib, which is very far from crypto grade. It's no problem, because this is not defined to be crypto grade.

I see your point. Perhaps there is a compromise somewhere? Providing the information can be useful when a knowledgeable user (perhaps another contributor, a downstream maintainer) runs into an issue, but you still have to make clear (somehow) that the implementation is subject to change. Perhaps a special indicator in the changelog to indicate whether the feature is stable or not?

I dunno, just a thought.

I'll go ahead and merge this.

deoren:
> @deoren: I'll go ahead and merge this.

I meant to say that I'll go ahead and merge this based off of your "looks better" remark.

davidelang:
On Thu, 25 Jan 2018, Deoren Moor wrote:

>> @rgerhards: Sometimes it is even bad -- because if we document it, we must support it in the future. For example, a PRNG crypto-grade gurantee would be a real problem on e.g. Solaris, where it is hard to obtain. Here we fall back to just the clib, which is very far from crypto grade. It's no problem, because this is not defined to be crypto grade.
>
> I see your point. Perhaps there is a compromise somewhere? Providing the information can be useful when a knowledgeable user (perhaps another contributor, a downstream maintainer) runs into an issue, but you still have to make clear (somehow) that the implementation is subject to change. Perhaps a special indicator in the changelog to indicate whether the feature is stable or not?

In this case, there should be no reason for it to ever be crypto-grade (and 
there are those who will argue that /dev/urandom isn't crypto-grade)

so it's best to avoid any inference that this is strong. It's designed for 
statistical sampling/balancing, nothing more.

davidelang:
On Sun, 4 Feb 2018, Deoren Moor wrote:

> @davidelang If I understand you, then even what we have now needs to be modified to remove all references to crypto. Are you suggesting something like this?
>
> ```diff
> -The random number must not be assumed to be crypto-grade. rsyslog will attempt to use a crypto-grade source if available, but it will fall back to a very weak random number generator if a crypto-grade source is not available.
> +The random number must not be assumed to be crypto-grade.
> ```

Yes, the rest of that is too much detail and locks us in.

For example, I could see the random number generator being something that's 
per-thread, and seeded off some something at boot time.

If you think of how early games asked for a random seed at startup, and the same 
seed generated the same sequence of numbers, that sort of very poor randomness 
is sufficient for the purposes of the random function (and for debugging could 
even be seen as a desirable thing)

On Linux /dev/urandom is fast enough and easy enough to use that implementing 
such an internal random number generator is more work, so we'll just use urandom 
:-) but on other systems, and in the future (such as the 4000 clock cycle 
penalty that spectre/meltdown protection can add to a system call), things can 
change

deoren:
@davidelang Fair enough. Thank you for taking the time to go into detail. I'll open a new PR for the changes we discussed here.

deoren:
@davidelang I submitted PR #539 for your review.

PR #511, opened 2018-01-21
Add standard footer

deoren:
> Instead of maintaining the substitution definitions within the Sphinx build configuration file, those definitions have been moved to a separate include file. Accordingly, the build conf has been updated to add an exclude entry for all *.inc.rst files so that they are not included multiple times during build
> processing. This include file is currently referenced by the rst_prolog build conf option.
> 
> A footer has been created and is included by all pages via an include statement within the rst_epilog build conf option. This footer is an attempt to display common "go here for help" options that should be made prominent for newcomers to the project.

@rgerhards What do you think about this idea? For the final merge, I would probably break up the two changes into either separate PRs or separate commits.

This will be the live view once the current build finishes (about 10 minutes after I post this):

http://rsyslog.whyaskwhy.org/standard_footer/

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -72,6 +72,14 @@ rst_prolog = """
 
 """
 
+# http://www.sphinx-doc.org/en/stable/config.html#confval-rst_epilog
+# This will be included at the end of every source file that is read.
+rst_epilog = """
+
+.. include:: /footer.inc.rst
+
+"""
+
 
 
 
--- a/source/footer.inc.rst
+++ b/source/footer.inc.rst
@@ -1,0 +1,16 @@
+
+
+.. seealso::
+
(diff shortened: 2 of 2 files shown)


rgerhards:
> @rgerhards What do you think about this idea? For the final merge, I would probably break up the two changes into either separate PRs or separate commits.

I like it very much!

deoren:
> @rgerhards: I like it very much!

Thanks! 

I'll work to split the changes out, but will probably wait to merge them until other PRs are merged first since those touch the files I will be working with here.

deoren:
Spun off PR #522 for moving substitution definitions to an external include file. I'm going to reuse this PR to handle adding the footer only. I'm rebasing off of master now and will merge this PR soon after.

Issue #489, opened 2018-01-09
source/rainerscript must be merged into configuration section


rgerhards:
It is laid out separately, which obviously does not match current state of the art. It now is an integral part of the configuration.

PR #512, opened 2018-01-21
describe new backtick string format

rgerhards:
... and also the others, as they were not yet described

see also https://github.com/rsyslog/rsyslog/pull/2419

### DIFF
--- a/source/rainerscript/constant_strings.rst
+++ b/source/rainerscript/constant_strings.rst
@@ -1,11 +1,16 @@
-Constant Strings
+String Constants
 ================
 
+String constants are necessary in any scripting language. The provide
+values that never change during rsyslog's run.
+
+Uses
+----
 String constants are necessary in many places: comparisons,
 configuration parameter values and function arguments, to name a few
 important ones.
 
-In constant strings, special characters are escape by prepending a
+In string constants, special characters are escaped by prepending a
 backslash in front of them -- just in the same way this is done in the C
 programming language or PHP.
 
@@ -13,3 +18,45 @@ If in doubt how to properly escape, use the `RainerScript String Escape
 Online
 Tool <http://www.rsyslog.com/rainerscript-constant-string-escaper/>`_.
 
+Types
+-----
+
+Rsyslog provides different types of string constants, closely inspired
+by the shell:
+
+- single quotes
+
+  Values are used unaltered, execept for escape sequences, which are
+  escaped.
+
+- double quotes
+
+  Right now, equivalent to single quotes, but $ signs need to be escaped.
+  If not escaped, a syntax error will be generated and rsyslog startup
+  be aborted in most cases.
+  The idea is to support environment variables just like the shell does
+  in later releases.
+
+- backticks
+
+  This was added in 8.33.0. The idea is to provide a useful subset of
+  what the shell does. Right now, only the following is supported:
+
+  - `echo $VARNAME` - It will evaluate the environment variable and use
+    it as string constant.  If the variable is not found, an empty string
+    is generated (this is **not** an error).
+
+  - `cat filename` - It will evaluate to the content of the given file.
+    Only a single file name is supported. If the file is not readable,
+    it will evaluate to an empty string.
+
+  Any other construct will currently lead to an error message.
+  Note that there must be exactly one space between "echo" or "cat" and
+  the other parameter.
+
+  Backticks are especially useful for configuration files that are
+  auto-generated but need to contain a small set of special functionality.
+
+  For an example of this in action, have a look at the rsyslog docker
+  appliance available at
+  https://github.com/rsyslog/rsyslog-docker/tree/master/appliance/alpine.


source/rainerscript/constant_strings.rst
```diff
@@ -13,3 +13,36 @@ If in doubt how to properly escape, use the `RainerScript String Escape
 Online
 Tool <http://www.rsyslog.com/rainerscript-constant-string-escaper/>`_.
 
+Type of String Constants
```
deoren: Would this be "Types of String Constants" instead?

Some other ideas (none of which are blockers):

1. Rename file to `string_constants.rst`
1. Update doc title to `String Constants`
1. Add `Uses` section with most of the info from the first section
1. Rename `Type of String Constants` to `Types`

I've push a branch to my repo that will build shortly for demo purposes.

source/rainerscript/constant_strings.rst
```diff
@@ -13,3 +13,36 @@ If in doubt how to properly escape, use the `RainerScript String Escape
...
+
+- single quotes
+
+  Values are used unaltered, execept for escape sequences, which are
+  escaped.
+
+- double quotes
+
+  Right now, equivalent to single quotes, but $ signs need to be escaped.
+  The idea is to support environment variables just like the shell does
```
deoren: What happens if `$` is not escaped?

deoren:
This is good stuff. I can already see where the new functionality will come in handy and having the existing behavior documented is also good. I've made a few in-line comments for consideration. I've also pushed a branch to my repo with some prototype changes for review. It should start building soon and I'll link to it here once it is ready.

deoren:
Link to live prototype:

http://rsyslog.whyaskwhy.org/pr512-prototype-string-contants/rainerscript/string_constants.html

rgerhards:
updated to now also include ``` `cat filename` ``` -- thought we should add this in the first shot ;-)

source/rainerscript/constant_strings.rst
```diff
@@ -1,15 +1,62 @@
...
+by the shell:
+
+- single quotes
+
+  Values are used unaltered, execept for escape sequences, which are
+  escaped.
+
+- double quotes
+
+  Right now, equivalent to single quotes, but $ signs need to be escaped.
```
deoren: What happens if `$` is not escaped?

deoren:
> @rgerhards: thx for the feedback!
>
> Updated with all except the file name. I think we should keep it stable until we (hopefully soon) integrate the ./rainerscript files into the main configuration section. Do not confuse Google twice ;-)

Alright. I see that you're referring to #489. If the `./rainerscript` files are merged into the main section then we could update the filename then as needed to match the surrounding content.

> @rgerhards: updated to now also include `cat filename` -- thought we should add this in the first shot ;-)

PRs rsyslog/rsyslog#2426 and rsyslog/rsyslog#2419 are game changers for a lot of configurations. I'm really looking forward to 8.33.0 now. :)

deoren:
Just in case it dropped out of sight, I had one question:

> Right now, equivalent to single quotes, but $ signs need to be escaped

What happens if the user does not escape the `$` sign?

rgerhards:
I thought I added the info on not escaped?

source/rainerscript/constant_strings.rst
rgerhards: Yup, next line 😛😉

deoren: Evidently I'm blind, I see it now:

> If not escaped, a syntax error will be generated and rsyslog startup be aborted in most cases.

 Thanks!

deoren:
Aside from the filename (which you already indicated would be handled on #489, it looks like this is ready to be merged.

source/rainerscript/constant_strings.rst
```diff
@@ -1,15 +1,62 @@
...
+values that never change during rsyslog's run.
+
+Uses
+----
 String constants are necessary in many places: comparisons,
 configuration parameter values and function arguments, to name a few
 important ones.
 
-In constant strings, special characters are escape by prepending a
+In string constants, special characters are escape by prepending a
```
deoren: I missed this in the last review:

```diff
-special characters are escape
+special characters are escaped
```

source/rainerscript/constant_strings.rst
```diff
@@ -1,15 +1,62 @@
...
 programming language or PHP.
 
 If in doubt how to properly escape, use the `RainerScript String Escape
 Online
 Tool <http://www.rsyslog.com/rainerscript-constant-string-escaper/>`_.
 
+Types
+-----
+
+Rsyslog provides different types of string constant, closely inspired
```
deoren:
```diff
-different types of string constant
+different types of string constants
```

rgerhards:
thx, fixed -- please merge if you think it is ok now

davidelang:
I'm assuming that this evaluation only happens at startup? This should be documentation.

deoren:
> @davidelang: I'm assuming that this evaluation only happens at startup? This should be documentation.

I went ahead and created a new issue for this (too easy to quickly glance at remarks on closed issues and then forget to circle back and complete the work). Perhaps not everyone has that problem, but I get distracted easily. ;)

rgerhards:
@davidelang just for ref, a quick answer: https://github.com/rsyslog/rsyslog-doc/issues/544#event-1459563817

Issue #502, opened 2018-01-12
Footnotes are busted on the main index


deoren:
The 'Sponsors and Community' section and 'Related Links' section have broken footnote references.

PR #514, opened 2018-01-23
Remove broken footnotes, 'Related Links' section

deoren:
The footnote references were broken and the 'Related Links' section was bare, so I pruned them and moved the Sponsors link just below the 'Sponsors and Community' section intentionally so it stands out.

Once some PRs ahead of this one goes through I plan to move the external link reference to
the substitution definitions file (again, in another PR) since this is probably not the only place we will reference the Sponsors link.

Additionally, I removed the explicit link text for the "How You Can Help" page since the title for that page is used if you do not override it with explicit link text.

The changes should be live here soon:

http://rsyslog.whyaskwhy.org/i502-footnote-fixes/

closes rsyslog/rsyslog-doc#502

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -53,20 +53,13 @@ Reference
 Sponsors and Community
 ----------------------
 
-Please visit the rsyslog Sponsor's Page\ [4]_ to honor the project sponsors or 
-become one yourself! We are very grateful for any help towards the project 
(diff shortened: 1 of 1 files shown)


deoren:
@rgerhards This seems like an easy one to judge, so I'm going to go ahead and merge it in. As always, please let me know whether I step over a line somewhere (as it won't be intentional).

rgerhards:
great work, thx!

PR #515, opened 2018-01-24
document the new 8.33 include() object

rgerhards:
The documents the new objects and its properties.

Note that I did deliberately not try to get to a "nice formatting", as
we are currently standardizing on something, and so a change is needed
in any case. Even more so, the whole (Rainer)Script documentation needs
to be restructured. So this patch gives the necessary information on
include(), but does not try to improve the state of the doc by itself.

see also https://github.com/rsyslog/rsyslog/pull/2426


deoren:
@rgerhards I'm really looking forward to using the new functionality. :)

PR #517, opened 2018-01-24
document 8.33 config object parameter "config.enable"

rgerhards:
see also https://github.com/rsyslog/rsyslog/pull/2433

### DIFF
--- a/source/rainerscript/configuration_objects.rst
+++ b/source/rainerscript/configuration_objects.rst
@@ -3,6 +3,33 @@ configuration objects
 
 Note: configuration object parameters are case-insensitive.
 
+Common Parameters
+-----------------
+
+config.enable
+.............
+
+.. versionadded:: 8.33.0
+
+All configuration objects have a parameter "config.enable".
+For auto-generated configs, it is useful to have the ability to disable some
+config constructs even though they may be specified inside the config. This
+can be done via the ``config.enabled`` parameter.
+If set to ``on`` or not specified, the construct will be
+used, if set to any other value, it will be ignored.
+This can be used together with the backtick functionality to enable or
+disable parts of the configuration from either a file or environment variable.
+
+Example:
+
+Let's say we want to conditionally load a module. Environment variable
+``LOAD_IMPTCP`` will be either unset or ``off`` .
+Then we can use this config construct::
+
+    module(load="imptcp" config.enabled=`echo $LOAD_IMPTCP`)
+
+If the variable is set to ```off```, the module will **not** be loaded.
+
 action()
 --------
 


source/rainerscript/configuration_objects.rst
```diff
@@ -3,6 +3,25 @@ configuration objects
...
+Example:
+
+Let's say we want to conditionally load a module. Environment variable
+``LOAD_IMPTCP`` will be either unset or ``off`` .
+Then we can use this config construct::
+
+    module(load="imptcp" config.enabled=`echo $LOAD_IMPTCP`)
+
+It the variable is set to ```off```, the module will **not** be loaded.
+
```
deoren: Suggestion: Add `.. versionadded: 8.33.0` here.

source/rainerscript/configuration_objects.rst
```diff
@@ -3,6 +3,25 @@ configuration objects
 
 Note: configuration object parameters are case-insensitive.
 
+All configuration objects have a parameter "config.enable".
+For auto-generated configs, it is useful to have the ability to disable some
+config constructs even though they may be specified inside the config. This
+can be done via the ``config.disable`` parameter.
```
deoren: Both `config.disable` and `config.enable` parameters are available?

source/rainerscript/configuration_objects.rst
```diff
@@ -3,6 +3,25 @@ configuration objects
...
+
+Example:
+
+Let's say we want to conditionally load a module. Environment variable
+``LOAD_IMPTCP`` will be either unset or ``off`` .
+Then we can use this config construct::
+
+    module(load="imptcp" config.enabled=`echo $LOAD_IMPTCP`)
+
+It the variable is set to ```off```, the module will **not** be loaded.
```
deoren: Typo:

```diff
-It
+If
```

> `module(load="imptcp" config.enabled=`echo $LOAD_IMPTCP`)`
>  the variable is set to ```off```, the module will **not** be loaded.

and if `config.enabled` is *not* explicitly set to `off` (either explicitly on or not specified) then the module is loaded?

source/rainerscript/configuration_objects.rst
```diff
@@ -3,6 +3,25 @@ configuration objects
 
 Note: configuration object parameters are case-insensitive.
 
+All configuration objects have a parameter "config.enable".
+For auto-generated configs, it is useful to have the ability to disable some
+config constructs even though they may be specified inside the config. This
+can be done via the ``config.disable`` parameter.
+If set to ``on`` or not specified, the construct will be
+used, if set to any other value, it will be ignored. This can be used
+together with the backtick functionality to configure enable and disable
```
deoren: > This can be used together with the backtick functionality to configure enable and disable from either a file or environment variable.

If I understand what this is saying, perhaps reword like so:

*This can be used together with the backtick functionality to enable or disable configure constructs from either a file or environment variable.*

In my mind, this altered wording places emphasis on the block of code that is toggled on or off by the file or environment variable.

I'm assuming here that "constructs" is the right term. I myself tend to be inconsistent between "blocks", "stanzas" and other related phrases.

source/rainerscript/configuration_objects.rst
rgerhards: no, I'll clear up. it's a boolean parameter so you have "on" and "off" values.

source/rainerscript/configuration_objects.rst
rgerhards: I have lightly changed it to *This can be used together with the backtick functionality to enable or
disable parts of the configuration from either a file or environment variable.*

source/rainerscript/configuration_objects.rst
rgerhards: I added it, but I do not see it causing any effect. Maybe for future use? Or does it need to be after the description? I needed to add a new section to make clear that this applies only to the parameter, not the config object itself. I then placed it right after the header (see updated commit).

deoren: Looks like I made a typo: there should be two `:` characters, so:

`.. versionadded:: 8.33.0`

Sorry about that.

deoren:
> Looks like I made a typo: there should be two : characters, so:
>
> .. versionadded:: 8.33.0
> 
> Sorry about that.

Reposting in case my earlier remark was missed when the older review comments were hidden.

rgerhards:
@deoren thx, that did the trick. I should also have noticed ;-)

PR #523, opened 2018-01-25
tooling: initial standard build container definition

rgerhards:
This is to be extended, but provides a first impression of how
this could be done. It's useful in any case to use during CI, as
it permits to reproduce exact same results in different environments.

To give the container a try, you need a system with docker installed. Then, run
```
$ docker run -v /path/to/rsyslog-doc_home_dir:/rsyslog-doc rsyslog/rsyslog_doc_gen
```
I have uploaded the container image to above docker hub location.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ source/_build
 .project
 .vscode
 *.pyc
+*.sw?
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
-sudo: false
-
-language: python
-
-python: "2.7"
+sudo: required
+services: docker
+language: minimal
 
 install:
-  - pip install -r requirements.txt
+  - docker pull rsyslog/rsyslog_doc_gen
 
 script:
-  - sphinx-build -n -W -b html source build
+  - docker run -e SPHINX_EXTRA_OPTS=-q
+        -v "$PWD":/rsyslog-doc -ti
+        rsyslog/rsyslog_doc_gen
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,0 +1,2 @@
+This directory contains tools primarily of use for maintainers
+or people who want to build the doc (on a schedule).
--- a/tools/buildenv/.dockerignore
+++ b/tools/buildenv/.dockerignore
@@ -1,0 +1,3 @@
+tmp*
+*.tmp
+README.*
--- a/tools/buildenv/Dockerfile
+++ b/tools/buildenv/Dockerfile
@@ -1,0 +1,17 @@
+FROM	alpine:3.7
+LABEL	maintainer rgerhards@adiscon.com
+RUN	apk add --no-cache py-pip git
+RUN	pip install sphinx
+RUN	adduser -s /bin/ash -D rsyslog rsyslog \
+	&& echo "rsyslog ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
+WORKDIR	/home/appliance
+CMD	["build-doc"]
+ENTRYPOINT ["/home/appliance/starter.sh"]
+VOLUME	/rsyslog-doc
+RUN	chmod a+w /rsyslog-doc
+ENV	BRANCH="v8-stable" \
+	FORMAT="html" \
+	STRICT="-n -W"
+COPY	starter.sh ./
+COPY	tools/* ./tools/
+# USER	rsyslog # does currentyl not play with VOLUME -- TODO: fix
--- a/tools/buildenv/README.md
+++ b/tools/buildenv/README.md
@@ -1,0 +1,6 @@
+build:
+
+docker build  -t rsyslog/rsyslog_doc_gen .
+
+where rsyslog/rsyslog_doc_gen is the tag and the location on dockerhub. Obviously,
+you need to use something else if you are not a maintainer of that repository.
--- a/tools/buildenv/starter.sh
+++ b/tools/buildenv/starter.sh
@@ -1,0 +1,9 @@
+#!/bin/ash
+if [ -f tools/$1 ]; then
+	source tools/$1
+else
+	echo "ERROR: command not known: $*"
+	echo
+	source tools/help
+	exit 1
+fi
--- a/tools/buildenv/tools/bash
+++ b/tools/buildenv/tools/bash
@@ -1,0 +1,2 @@
+echo "You are now in Alpine's ash"
+/bin/ash
--- a/tools/buildenv/tools/build-doc
+++ b/tools/buildenv/tools/build-doc
@@ -1,0 +1,22 @@
+if [ ! -e /rsyslog-doc/source ]; then
+	if [ -e /output ]; then
+		source tools/git-clone
+	else
+		echo "ERROR: /rsyslog-doc is either not mounted or does not point to correct path"
+		echo "       and /output is also not present, so we do not know what to do"
+		source tools/help
+		exit 1
+	fi
+fi
+if [ -e /output ]; then
+	export OUTPUT=/output
+else
+	export OUTPUT=build
+fi
+cd /rsyslog-doc
+sphinx-build $STRICT -b $FORMAT $SPHINX_EXTRA_OPTS source $OUTPUT
+RESULT=$?
+# we need to do the chown as we currently cannot select the right user
+# inside the Dockerfile. Once this is solved, this can go away.
+# chown -R rsyslog:rsyslog $OUTPUT - need to do this differently...
+exit $RESULT
--- a/tools/buildenv/tools/git-clone
+++ b/tools/buildenv/tools/git-clone
@@ -1,0 +1,13 @@
+if [ ! -e /rsyslog-doc ]; then
+	echo "ERROR: /rsyslog-doc is either not mounted or does not point to correct path"
+	source tools/help
+	exit 1
+fi
+if [ -e /rsyslog-doc/source ]; then
+	echo "ERROR: /rsyslog-doc contains a source directory, so we cannot clone into it"
+	source tools/help
+	exit 1
+fi
+cd /
+git clone https://github.com/rsyslog/rsyslog-doc
+(cd /rsyslog-doc; git checkout $BRANCH)
--- a/tools/buildenv/tools/help
+++ b/tools/buildenv/tools/help
@@ -1,0 +1,20 @@
+echo 'docker run -v /your/local/path:/rsyslog-doc <container-name> command
+where command is one of:
+build-doc    - generate documentation (currently HTML only)
+help         - display this help message
+bash         - start an interactive shell (<CTL>-d terminates)
+version-info - outputs some information of component used in container
+git-clone    - clone the official rsyslog-doc project into /rsyslog mount
+               this is a quick way to populate /rsyslog, which must be empty before
+
+Environment variables (set with -eVAR=val on docker run):
+FORMAT       - html (default) or another builder format
+STRICT       - options to cause strict sphinx mode (default "-n -W")
+SPHINX_EXTRA_OPTS - provides any option to sphinx the user whiches, e.g. -q for quiet builds
+
+bind mounts (mount with -v/local/path:/container on docker run):
+/output      - output files will be placed here. If not mounted, output will be under
+               /rsyslog-doc/output
+/rsyslog-doc - checked-out rsyslog doc project; if not mounted but /output is, an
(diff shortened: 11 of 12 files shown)


deoren:
I've installed docker on an Ubuntu 16.04 VM by following the directions here:

https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/

I then ran:

1. `cd Desktop`
1. `sudo apt-get install git-core`
1. `git clone https://github.com/rgerhards/rsyslog-doc`
1. `git checkout docker`
1. `docker run -v $PWD:/rsyslog-doc rsyslog/rsyslog_doc_gen`
    - the last bit I got an error re permissions, so I ran again with sudo
1. `sudo docker run -v $PWD:/rsyslog-doc rsyslog/rsyslog_doc_gen`

The docs were then built (apparently `master`) and control was returned to my current shell, presumably outside of docker. From what I can tell the entire process acted like a "one-shot" script.

This was more involved than running the pip commands to setup and build the docs as a non-root user, but it was self-contained, so that's interesting.

rgerhards:
It's most probably not a replacement if you have python installed but not docker. Cons IMHO are:

- it's always the same environment
  That's especially good under CI (maybe a bit less with sphinx as there are not so big differences). 
  With some tools, it really drives one crazy if you can't reproduce the errors/warning locally that
  CI has...

- you do not need to care about changing anything on your system
  (assuming you have docker installed)

- the production machine can use the same script

- of course some bits are missing, just have added a checkout capability. Next should be the ability
  to run the web update script automatically.

I have also amended the build so that travis uses it. Let's see how this comes out.

rgerhards:
oh, I forgot: the "sudo-thing" is well known and usually adressed one way or the other if you run docker.

deoren:
This is very promising. Looks like I'll have to learn docker sooner than I had planned, but oh well, this is the field we work in. :)

rgerhards:
> This is very promising. Looks like I'll have to learn docker sooner than I had planned, but oh well, this is the field we work in. :)

It's not as bad as it initially sounds. I begun just recently, and if you created VM setup automation, it's really nothing new. Look at the Dockerfile inside the PR and then check ```starter.sh```. The later is the process that is automatically started and then passed any commands (with ```CMD``` being the default).

deoren:
> @rgerhards: oh, I forgot: the "sudo-thing" is well known and usually adressed one way or the other if you run docker.

I'm guessing that it involves adding the user to a local group or some such. I'll look into that.

rgerhards:
> I'm guessing that it involves adding the user to a local group or some such. I'll look into that.

The short answer is "yes" (docker IIRC), the longer is that it comes with some security risk. I've done it that way on a couple of dev VMs, but not all.

rgerhards:
Oh, and I overlooked this: step 1-4 were actually **not** necessary - ```docker run``` is all you need. It takes the image from dockerhub (think of it as a kind of github for the container images, or as a package repository). It's there because I uploaded it.

deoren:
> @rgerhards: Oh, and I overlooked this: step 1-4 were actually not necessary - docker run is all you need. It takes the image from dockerhub (think of it as a kind of github for the container images, or as a package repository). It's there because I uploaded it.

I've not been able to get the image to start without those steps. I'm sure it's something I'm doing wrong, but the command seems to expect that I'm currently in the working area for the repo, or at least my current working directory is where the source directory can be found.

rgerhards:
> I've not been able to get the image to start without those steps. I'm sure it's something I'm doing wrong, but the command seems to expect that I'm currently in the working area for the repo, or at least my current working directory is where the source directory can be found.

ah, ok, my fault -- I assumed you were in a working copy. In the mean time I have also updated the container so that it can do the git pull itself. You still need to have a host directory mapped (we remove that requirement), so the steps then are:

* ``docker run -v /your/dir:/rsyslog-doc -ti rgerhards/rsyslog_doc_gen git-checkout```
* ``docker run -v /your/dir:/rsyslog-doc -ti rgerhards/rsyslog_doc_gen```

That's possible because git is present inside the container.

rgerhards:
Further note: I updated the Travis definitions, CI now uses the container. It uses a two-step process, where the image is downloaded (```docker pull```) at ```install``` time. This is not strictly necessary, but useful with Travis so that a network error is more clearly indicted inside the status display (exclamation mark symbol).

deoren:
> @rgerhards: ah, ok, my fault -- I assumed you were in a working copy. 

I used a fresh VM for the test so I could see what steps were needed from fresh boot, to docker install to loading the docker image you supplied.

> @rgerhards: You still need to have a host directory mapped

:) I'm *that* new to docker; before today I've never installed or booted a docker image. I came to that conclusion though looking at the error output when I tried shortening the `docker run` command even further than it already was.

> @rgerhards: Further note: I updated the Travis definitions, CI now uses the container. It uses a two-step process, where the image is downloaded (docker pull) at install time. This is not strictly necessary, but useful with Travis so that a network error is more clearly indicted inside the status display (exclamation mark symbol).

Very nice. I'm impressed that Travis supports that approach.

On a tangent, but is there any concern re bandwidth usage of a full clone/build vs having a mirror of the repo available to clone locally for each build?

In other words, if on a build box (or the docker image I guess) a local mirror is available then a clone could be made of that mirror instead of directly from the remote. I'm testing that approach in the cron script so that it isn't constantly cloning directly from github.com.

Perhaps I'm thinking about this the wrong way. For a one-off infrequent build, then I imagine cloning directly from github.com isn't a problem, but for automated builds I would imagine that the bandwidth load would be quite high.

rgerhards:
I'll elaborate more when I am back at a real keyboard, but (below)...

Sent from phone, thus brief.

Am 25.01.2018 19:28 schrieb "Deoren Moor" <notifications@github.com>:

@rgerhards <https://github.com/rgerhards>: ah, ok, my fault -- I assumed
you were in a working copy.

I used a fresh VM for the test so I could see what steps were needed from
fresh boot, to docker install to loading the docker image you supplied.

@rgerhards <https://github.com/rgerhards>: You still need to have a host
directory mapped

:) I'm *that* new to docker; before today I've never installed or booted a
docker image. I came to that conclusion though looking at the error output
when I tried shortening the docker run command even further than it already
was.


What the -v does is actually a Unix mount, so the data resides on the
"host" system (that runs the container).

@rgerhards <https://github.com/rgerhards>: Further note: I updated the
Travis definitions, CI now uses the container. It uses a two-step process,
where the image is downloaded (docker pull) at install time. This is not
strictly necessary, but useful with Travis so that a network error is more
clearly indicted inside the status display (exclamation mark symbol).

Very nice. I'm impressed that Travis supports that approach.

On a tangent, but is there any concern re bandwidth usage of a full
clone/build vs having a mirror of the repo available to clone locally for
each build?


Travis: never - you always get a fresh VM.

In general: not for projects of our size. For CI it is even important to
start as fresh as possible. The buildbot machines aldi clean out the local
copy iIRC.

In other words, if on a build box (or the docker image I guess)


You usually should not have the data on a docker image but on the host
machine (or some Network store). Docker container != VM. A container is
much more like an app than a vm. I would think about it as another way to
link a binary (ok, that is also a very extreme way of thinking, but not as
wrong as it initially looks).

a local mirror is available then a clone could be made of that mirror
instead of directly from the remote. I'm testing that approach in the cron
script so that it isn't constantly cloning directly from github.com.


TBH I would not bother. Even more so as the infrastructure more and more
spawns short lived VMs and containers.

However, one clone should be sufficient to handle all branches.

Perhaps I'm thinking about this the wrong way. For a one-off infrequent
build, then I imagine cloning directly from github.com isn't a problem, but
for automated builds I would imagine that the bandwidth load would be quite
high.


Where do you see the bandwidth consumption?

deoren:
> @rgerhards: What the -v does is actually a Unix mount, so the data resides on the
"host" system (that runs the container).

Gotcha, that makes sense.

> > @deoren: On a tangent, but is there any concern re bandwidth usage of a full
clone/build vs having a mirror of the repo available to clone locally for
each build?
>
> @rgerhards: Travis: never - you always get a fresh VM.
>
> In general: not for projects of our size. For CI it is even important to
> start as fresh as possible. The buildbot machines aldi clean out the local
> copy iIRC.
> 
> In other words, if on a build box (or the docker image I guess)
> 
> You usually should not have the data on a docker image but on the host
> machine (or some Network store). Docker container != VM. A container is
> much more like an app than a vm. I would think about it as another way to
> link a binary (ok, that is also a very extreme way of thinking, but not as
> wrong as it initially looks).

Alright. So deploying a clone of the repo would be a bad idea, but having one outside of the docker image and then mounting it within could work.

> > @deoren: a local mirror is available then a clone could be made of that mirror
instead of directly from the remote. I'm testing that approach in the cron
script so that it isn't constantly cloning directly from github.com.
>
> @rgerhards: 
> TBH I would not bother. Even more so as the infrastructure more and more
spawns short lived VMs and containers.
> 
> However, one clone should be sufficient to handle all branches.

By "one clone" you mean one clone operation before building is all that is needed? If so I agree. My thought was that I could keep a mirror (`git clone --mirror https://github.com/deoren/rsyslog-doc rsyslog-doc-mirror.git`) locally that is synced on each script run and used as a source when running `git clone` (e.g., `git clone /path/to/rsyslog-doc-mirror.git rsyslog-doc`). That way if I crank up the frequency for builds it wouldn't increased the bandwidth usage by a whole lot. 

>> @deoren: Perhaps I'm thinking about this the wrong way. For a one-off infrequent
build, then I imagine cloning directly from github.com isn't a problem, but
for automated builds I would imagine that the bandwidth load would be quite
high.
>
> @rgerhards: Where do you see the bandwidth consumption?

I don't see it exposed anywhere (I have not looked), but if you look at what I'm doing now (which is admittedly inefficient) and then move to performing a fresh clone each time the script runs:

1. Cron script executes
1. Fresh clone is made
1. Branches are built
1. Clone is toss
1. Repeat

I would expect that with a frequent build rate it would eventually chew up some bandwidth. Maybe not enough to matter, probably not even enough to matter on my VPS (seems they rate by traffic going out, not in), but it just seems rude to hit github.com with regular clone requests that aren't necessary. That said, I still vividly recall dial-up modem access days, so "inefficient" may not really apply anymore.

If Travis CI and others CI platforms are not providing any content caching and each build they perform results in a fresh clone, then perhaps github.com is so well built out that there is just not an issue, no matter how many builds that the community at large runs.

rgerhards:
> By "one clone" you mean one clone operation before building is all that is needed?

It of course all depends on how you build. The container is not necessarily different than a "local" sphinx build. If all else is present, just think about it as another way to run sphinx-build.


> I would expect that with a frequent build rate it would eventually chew up some bandwidth. Maybe not enough to matter, probably not even enough to matter on my VPS (seems they rate by traffic going out, not in), but it just seems rude to hit github.com with regular clone requests that aren't necessary. That said, I still vividly recall dial-up modem access days, so "inefficient" may not really apply anymore.

IMHO "the world has changed". Let's put it into perspective: cloning the rsyslog-doc project is roughly 4MiB traffic (3.5MiB reported by git as I did it just now + some overhead). One second of serving HD video  at 1080p is ~22Mbps --> ~2.8MiB. So a git clone is roughly equivalent to 1.5 seconds of video streaming. 

> If Travis CI and others CI platforms are not providing any content caching and each build they perform results in a fresh clone, then perhaps github.com is so well built out that there is just not an issue, no matter how many builds that the community at large runs.

I think it really doesn't matter to them. I also would assume that, as almost all folks do, they run on AWS and/or GCE, so that's well handled. I guess that the regular DDoS traffic outperforms the overall clone rate.

That said, fair use still matters. I would say it is no problem if you do a full clone every hour or so, but doing it every minute I would consider abusive. As part of CI, of course, more frequent clones are perfectly valid. E.g. every PR you do on the rsyslog main project results in at least 25 clone operations, because we run CI on that many platforms.

For the doc, I would still set up a single persistent repro (possibly ;-)) that would just to be ```git pull```'ed on each invokation. The current script does do that.

rgerhards:
@deoren I am now going ahead and merge this, as it does not change the doc itself and I have thoroughly checked that it works. Also, I have changed the web update script to use the container, so it would be good to have it's definition inside the repository.

I plan to do some more work on the container later on.

deoren:
@rgerhards Acknowledged, thanks for the detailed feedback.

I muddied the water a bit by mixing in concerns with regular CI runs (git clone operations) and regular cron builds (also git clone operations). While the two are related in that content is frequently fetched, the goals are different (at least IMO).

I'll discuss further on the experimental branch ticket instead of here.

Issue #524, opened 2018-01-26
cron_build_all_branches does not work on Alpine Linux


rgerhards:
* it requires bash, which is not enabled by default on Alpine
* if bash is installed, it fails on some sort option, albeit the rest of the build seems to succeed, but...
* it fails at the end as follows:
```
build succeeded.
tar: build.sh: No such file or directory
tar: error exit delayed from previous errors
[!] tarball creation failed for  tag ... aborting
/rsyslog-doc # 
```

Also, during build there are many warnings, which otherwise do not appear when the branches are build. Not sure if that is correct or an Alpine-induced problem.

related: https://github.com/rsyslog/rsyslog-doc/pull/523

deoren:
@rgerhards thanks for the report. 

> `tar: build.sh: No such file or directory`

I'm guessing this is for the v8-stable branch?

The v8-stable copy of the script still references the `build.sh` script:

`tar -czf $output_dir/$doc_tarball source build LICENSE README.md build.sh`

Current master does not:

`tar -czf $output_dir/$doc_tarball source build LICENSE README.md`

When I first included that entry the idea was that the stable release tarballs could be automated so that when you got ready to release one you could just fetch it from the build system and deploy it. I've realized that building tarballs via the cron script is probably just a waste of time and lengthens the build process.

deoren:
> if bash is installed, it fails on some sort option, albeit the rest of the build seems to succeed, but...

I wonder if the sort binary is provided by busybox? From what I recall the `sort` call is done to line up the git tags in order so that the latest one can be plucked from the list. I may be able to reliably get the latest tag available through another means without falling back to `sort` to achieve it.

`~/dev/rsyslog/rsyslog-doc$ git describe --tags`
```
v8.32.0-54-g315fcde
```

That *seems* to work, but I've not tested it in practice to determine whether it's reliable.

rgerhards:
@deoren I just wanted to make you aware of the issue. I think it can probably simply be "resolved" by documenting it. I tend to think that most folks will probably need to add some tweaks to cron scripts, so I am not sure how frequently it will be used.

deoren:
> @rgerhards: Also, during build there are many warnings, which otherwise do not appear when the branches are build. Not sure if that is correct or an Alpine-induced problem.

I'll try to replicate that using the docker setup you created for the docs. If nothing else it will give me some additional practice using docker.

> @rgerhards: I just wanted to make you aware of the issue. 

[...]

PR #526, opened 2018-01-26
cron script: mimic '--version-sort' option for busybox compatibility

deoren:
Instead of using the GNU sort utility's "version sort" option, fall back to a combination of options that appear to provide the same effect, but also work well with busybox (used in some docker container images).

refs rsyslog/rsyslog-doc#524


deoren:
@rgerhards I've used this version of the script inside of the `rsyslog/rsyslog_doc_gen` image and I no longer see any errors related to the build script. Going to go ahead and merge to master to get the fix in.

PR #527, opened 2018-01-27
Dockerfile: Change default branch to build from v8-stable to master

deoren:
I see where you're checking first to make sure that they don't already have a bind mount for a checked out version of the docs. If the docs are found, they're built as-is without changing the checked out branch. If they don't already have a copy of the docs, then a git clone operation occurs, a checkout call and then a build of the checked out branch. That makes sense. You're essentially emulating someone calling `sphinx-build -b html source build` directly against content in the current working directory.

Provided that the docker image is not meant to be dedicated to building the stable version of the docs, I suggest favoring `master` over `v8-stable` as that fallback branch choice is likely to be of greater interest  to someone opting to build the docs on their own.

### DIFF
--- a/tools/buildenv/Dockerfile
+++ b/tools/buildenv/Dockerfile
@@ -9,7 +9,7 @@ CMD	["build-doc"]
 ENTRYPOINT ["/home/appliance/starter.sh"]
 VOLUME	/rsyslog-doc
 RUN	chmod a+w /rsyslog-doc
-ENV	BRANCH="v8-stable" \
+ENV	BRANCH="master" \
 	FORMAT="html" \
 	STRICT="-n -W"
 COPY	starter.sh ./


rgerhards:
actually, I first had it default to ```master``` and only later changed it to ```v8-stable``` assuming that this would be the most sensible choice. I am actually a bit undecided, as both versions are available on the web site. But you have a point, probably the interest in ```master``` is greater. If you still agree, I think we should merge.

Side-note: you can override the environment variable on docker run via the ```-e``` option:
```
docker run -e BRANCH=v8.30.0 ...
```
will build the doc for v8.30.0.

deoren:
@rgerhards 

I'll go ahead and merge, but we can always change back or to something else later. Since the default action is to build the current content, that should cover most use cases.

deoren:
> @rgerhards: Side-note: you can override the environment variable on docker run via the -e option:
> 
> `docker run -e BRANCH=v8.30.0 ...`
> 
> will build the doc for v8.30.0.

That's useful. 

Tangent: 

In my readings I keep coming across examples of overriding behavior within containers using environment variables. I've not yet read far enough to learn about the security aspect of using environment variables for things like usernames or passwords. Presumably you have as few users as possible attaching to the image directly where they could read the environment variable values, so perhaps it's not a big issue.

Here is makes perfect sense to do what you've done as it is completely unrelated to security.

rgerhards:
> In my readings I keep coming across examples of overriding behavior within containers using environment variables. I've not yet read far enough to learn about the security aspect of using environment variables for things like usernames or passwords. Presumably you have as few users as possible attaching to the image directly where they could read the environment variable values, so perhaps it's not a big issue.

Couple of things:

1. whatever you store inside a container (e.g. usernames/passwords), it is always possible to access thaat information one way or the other for  someone who has access to the (container) image
2. if you set environment variables via docker run, you do so **outside of the container**. So anyone interested to learn their value must either
   - have access to the host environment directly (problem with CI runs, depending on setup)
   - must be able to place a malicious app into the container (making the container itself malicious)

My personal conclusion:
- passing credentials via environment variable (or a mount point) is probably the most secure way to do it
- ensure that you can trust the container!

Practical sample:

Have a look how I build Alpine packages. At the bottom step, I do:
    https://github.com/rsyslog/rsyslog-pkg-alpine/blob/master/run

Note that it passes the signature keys in via a bind mount (the alternative to environment variables). So nobody is able to obtain them just by having the container image.

deoren:
> @rgerhards: if you set environment variables via docker run, you do so outside of the container.

Ah, I didn't realize that. I assumed that the env variables (the setting of them) would be logged in the shell history file (what commands were run), but that they would be directly exposed only to the inside of the container.

> @rgerhards: Have a look how I build Alpine packages. At the bottom step, I do:
> https://github.com/rsyslog/rsyslog-pkg-alpine/blob/master/run
> 
> Note that it passes the signature keys in via a bind mount (the alternative to environment variables). So nobody is able to obtain them just by having the container image

Thank you for the example. I will make sure to refer back to the work you've done as I get closer to actually using docker on a regular basis. I appreciate your advice and feedback.

rgerhards:
> Ah, I didn't realize that. I assumed that the env variables (the setting of them) would be logged in the shell history file (what commands were run), but that they would be directly exposed only to the inside of the container.

good point - they will be recorded via the docker run, but outside of the container. So this environment must be sufficiently secure. Inside the container you don't have it in history (but of course you have the value!). Just done so:
```
rg@rg-virtual-machine:~/rsyslog-pkg-alpine$ docker run -ti --rm -eTEST=bla alpine
/ # env
HOSTNAME=ff913bf70c20
SHLVL=1
HOME=/root
TERM=xterm
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
TEST=bla
PWD=/
/ # history
   0 env
   1 history
/ # 
/ # rg@rg-virtual-machine:~/rsyslog-pkg-alpine$ history |tail -2
 2351  docker run -ti --rm -eTEST=bla alpine
 2352  history |tail -2
```
With a file:
```
rg@rg-virtual-machine:~/rsyslog-pkg-alpine$ cat /tmp/secret
uid
rg@rg-virtual-machine:~/rsyslog-pkg-alpine$ docker run -ti --rm -v/tmp/secret:/tmp/secret alpine
/ # cat /tmp/secret
uid
/ # history
   0 cat /tmp/secret
   1 history
/ # 
```
But this boils down to the usual problem of keeping credentials secure on a box.

rgerhards:
But be careful with work files and similar things inside the docker container:
```
:docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
:docker run -ti  -eTEST=bla alpine
/ # echo $TEST
bla
/ # echo $TEST > workfile
/ # :docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS                     PORTS               NAMES
a750e1833c38        alpine              "/bin/sh"           52 seconds ago      Exited (0) 7 seconds ago                       pensive_jepsen
:docker start pensive_jepsen
pensive_jepsen
:docker exec pensive_jepsen /bin/ash
ls
^C
:docker exec -ti pensive_jepsen /bin/ash
/ # ls
bin       etc       lib       mnt       root      sbin      sys       usr       workfile
dev       home      media     proc      run       srv       tmp       var
/ # cat workfile
bla
/ # :
```
Note: ```PS1=:``` set to keep the output readable.

Here I use the not-yet-discarded interim layer to obtain the workfile and thus the secret! Same goes for similar info as volume content. So you need to be pretty careful. Running the container read-only (if supported) helps with temp work files, of course. That's among others why I try to keep the rsyslog logging appliance as read-only except for the volumes.

PR #531, opened 2018-01-30
Add Table of Contents to README files

deoren:
Used Atom Editor + mardown-toc-auto plugin to generate them

https://atom.io/packages/markdown-toc-auto

Even so, it's quite possible to hand-edit the TOC entries should someone wish to note use the editor + plugin.

Thoughts?

Live view:

- https://github.com/deoren/rsyslog-doc/tree/add-readme-toc/README.md
- https://github.com/deoren/rsyslog-doc/blob/add-readme-toc/BUILDS_README.md

If this looks good, I'll followp-up with doing the same for the remaining README files.

### DIFF
--- a/BUILDS_README.md
+++ b/BUILDS_README.md
@@ -1,3 +1,17 @@
+<!-- TOC START min:1 max:3 link:true update:true -->
+- [rsyslog-doc builds: A guide to building the rsyslog documentation for maintainers and contributors](#rsyslog-doc-builds-a-guide-to-building-the-rsyslog-documentation-for-maintainers-and-contributors)
+  - [Generating release version of the docs](#generating-release-version-of-the-docs)
+    - [Maintainer](#maintainer)
+    - [Contributors](#contributors)
+  - [Generating development builds](#generating-development-builds)
+    - [Obtain docs](#obtain-docs)
+    - [Build from GitHub download](#build-from-github-download-1)
+    - [Build from Git repo](#build-from-git-repo-1)
+
+<!-- TOC END -->
+
+
+
 # rsyslog-doc builds: A guide to building the rsyslog documentation for maintainers and contributors
 
 *Documentation for building the documentation for ...*
--- a/README.md
+++ b/README.md
@@ -1,3 +1,20 @@
+<!-- TOC START min:1 max:3 link:true update:true -->
+- [rsyslog-docs](#rsyslog-docs)
+  - [Documentation for the rsyslog project](#documentation-for-the-rsyslog-project)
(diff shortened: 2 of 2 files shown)


rgerhards:
gut feeling only: I think we should not put too much into a single README file. Maybe it's better placed inside the main doc. Again, gut feeling after 30sec of thought... (wanted to give feedback today).

deoren:
> @rgerhards: gut feeling only: I think we should not put too much into a single README file.

In other words, if the README begins to grow so long that a TOC would be helpful, don't add a TOC as a finding aid, instead split the README file so that the main file acts as an index to the other material.

I considered that. I think it's a good idea that once a doc hits a certain size that it should be split out to make it easier to maintain.  I wasn't sure if we had hit that point yet, but I guess if I'm proposing a TOC we may have.

> @rgerhards: Maybe it's better placed inside the main doc. 

Perhaps. A good first step in either directions it to probably split out the building instructions from the main README, look over the second doc I wrote and see if the two can be easily merged. Once that is done we'll have a better idea of what we're working with and can decide on where it is housed: separate Markdown file in the main project root (for easy viewing from GitHub web UI) or inside the main doc (and linked from the main README file).

deoren:
@rgerhards

After giving this some thought I think having a ToC on the main README is probably a bad idea. Instead it should contain just the bare essentials and then forward the reader to other pages for additional coverage.

Currently we have a doc for dev builds and a main README with coverage of "standard" builds. I will work to place all build-related directions onto a single page. From that point it's not a lot of work to move them into the main docs if that is where we decide that is the best place for them.

I'll go ahead and close this out.

rgerhards:
We probably need to think about a general policy what should go into the doc and what not. That's most probably a discussion we should try to have on the mailing list, or otherwise try to gather as much feedback as possible.

I have a personal tendency to have everything in a single doc set, but that may be just me.

rgerhards:
I would generally say: go ahead, but with one warning:

> I see the actual documentation files as more for rsyslog itself.

I think this would benefit from a broader community discussion (if we manage to initiate a real one...). I myself (as a contributor) do lean towards the other direction, "put everything into the doc", but README outside doc is also good. I am not sure if README.md files are the right choice -- I do not want to get too locked in into github. Granted, it is a reputable service, but sourceforge.net was also a very good service for many years ... before it turned down. I was very happy we were not locked in to it ;-)

Again, speaking as contributor here. But I think it is an so important decision that I would like to hear at least some other votes from the community.

With that warning, I would be absolutely happy if you could go ahead as suggested. But for that you would need to be comfortable with possibly needing to change it to some other method in the future.

PR #533, opened 2018-02-01
use theme that integrates better with rsyslog.com site theme

rgerhards:


rgerhards:
@deoren the theme works much better on the rsyslog site, especially with small screens. I have manually generated an update to the master branch doc, have a look: http://www.rsyslog.com/doc/master/

I think this solves the problems we had with site integration, and does so mostly satisfactory. 

Unfortunately, I need a non-standard them (aka "needs to be installed via pip"). I want to avoid this, as I really think we cannot and should not put that burden on the user. I orignally thought I create an include file for the theme selection and include that in the main ```conf.py```. But I seem to miss the obvious ... can't find how to do that in python. Can you help?

If we have an include file, I can just overwrite that for site generation. If that doesn't work, I can ```sed``` the conf.py, but that doesn't sound like a good clean solution to me.

rgerhards:
I forgot: automatic site doc updates are currently disabled as I want to preserve the manually generated content. I think we can solve this within a day, so that should not be an issue.

deoren:
In no particular order, here is some feedback. I hope it's not harder to follow, but I will break my response up into several posts to help me focus.

First, overall I like the look of the new theme. It feels very clean and has a lot of potential for further styling (if needed/desired).

Regarding mobile support, I find mixed results, but I didn't extensively test before with the previous theme so it's hard to say whether this is expected or not.

- In Google Chrome or Mozilla Firefox for Android on a phone I see the content shifted to the left with plenty of "gray space" to the right. In landscape mode the content fills the screen.
- On a tablet (7"), I see the center column using all available space to center the content nicely. This is with Mozilla Firefox or Google Chrome.

I'm not a web developer, but I would guess that *something* is performing a screen size calculation to determine how much of the screen to use and oddly enough the larger screen (but perhaps lower resolution?) is getting the better fit.

On a different note, the sidebar links for the module section headers are a bit squished. See the "Purpose" and "Notable Features" sections:

imuxsock-module-better-theme-sphinx-2018-02-01

Not a deal breaker of any kind, but something that may need to be worked around in the surrounding content container (CSS styles?) at some point.

deoren:
## LG V20 Android phone: Google Chrome

screenshot_2018-02-01-23-23-44

## Nexus 7 tablet: Google Chrome

screenshot_20180201-232715

## Nexus 7 tablet: Mozilla Firefox

screenshot_20180201-232749

screenshot_20180201-233112

deoren:
> @rgerhards: I forgot: automatic site doc updates are currently disabled as I want to preserve the manually generated content. I think we can solve this within a day, so that should not be an issue.

I think so too. We're not pushing so many new changes these last few days that the live view of master will suffer much. :)

> @rgerhards: I think this solves the problems we had with site integration, and does so mostly satisfactory.

Yeah, the new theme looks pretty good when integrated, especially with the banner moved to the top. I suspect having the banner at the top makes the site more mobile friendly regardless of theme.

> @rgerhards: Unfortunately, I need a non-standard them (aka "needs to be installed via pip"). I want to avoid this, as I really think we cannot and should not put that burden on the user.

I think I mentioned some options via email when we spoke earlier, but I think we can rely on the somewhat recent addition of the `requirements.txt` file into the root of the project. Instead of directly referencing the `sphinx` package, our main [README](rsyslog/rsyslog-doc/blob/master/README.md) file now directs readers to run `pip install -r requirements.txt` to install sphinx. This has the rather nice side-effect (or in our case, intentional effect) of providing a way to easily expand package dependencies without complicating the setup/build process. We just update that file to include the `sphinx-better-theme` package and users that follow the directions (and any downstream package maintainers) to use the provided `requirements.txt` file will get the intended results.

Had the `sphinx-better-theme` package not been available through `pip`, then we could have looked into bundling the theme directly into the `source` folder. This would have resulted in the maintenance burden, however light, of keeping the bundled theme content up to date with upstream.

deoren:
> @rgerhards: I orignally thought I create an include file for the theme selection and include that in the main conf.py. But I seem to miss the obvious ... can't find how to do that in python. Can you help?
> 
> If we have an include file, I can just overwrite that for site generation. If that doesn't work, I can sed the conf.py, but that doesn't sound like a good clean solution to me.

I think what you're getting at is yet another option: conditional builds. We're doing some of that now with the dev vs release build logic which is implemented in `source/conf.py` (and recently merged to master. 

- Is the placeholder version still in place?
    - Yes: let's try to generate a dev build.
        - Is the `.git` subdirectory present?
             - Yes: let's use it to generate a dev release string and set a tag for conditional inclusion of a "dev" footer string on the main index page
             - No: let's just keep the placeholder "dev build" suffix; this is better than nothing
    - No: we have a stable release string, let's use that. Skip all other build-related checks.

With this scenario we could:

1. Set a variable to reference the desired theme
1. Try to import the desired theme
1. If import failure, set variable to default theme
1. Later in the file, set the `html_theme` option to reference the variable instead of a hard-coded theme name (which is the normal way to handle this).

That should provide a fallback for the build if someone doesn't have the desired theme package already installed.

source/conf.py
```diff
@@ -209,17 +209,22 @@
...
+#html_theme = 'default'
 #html_theme = 'basic'
 #html_style = 'rsyslog.css'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-#html_theme_options = {}
+html_theme_options = {
+'inlinecss': ' @media (max-width: 820px) { div.sphinxsidebar { visibility: hidden; } }',
```
deoren: I think this may be related to the resolution check that I mentioned in an earlier reply. I hadn't yet looked over the diff against the `conf.py` file yet when I responded.

It may not matter here, but for reference here are the resolution specs on the two devices I mentioned as my test cases earlier:

- LG V20 resolution is 2,560 x 1,440
- Nexus 7 (2013) resolution is 1920 x 1200

deoren:
@rgerhards See b57c0e4c0ea79652b41aa8cf09398959d7e25da4 for how to make the chosen theme preferred (but optional) with a fallback to the default theme should it not be present. If (as with that revision) you omit including the `sphinx-better-theme` package name in `requirements.txt`, then (for the most part) systems other than your buildbox (container or not) will generate the docs with the standard/default theme.

If you're fine with changing the theme for everyone (well, everyone that follows the guidance in the [README](/rsyslog/rsyslog-doc/blob/master/README.md file), then 67cfd32a3997923570092e3925188701e1b28358 illustrates the necessary change (ignoring the comments) to the `requirements.txt` file.

source/conf.py
rgerhards: that max width is from the better theme. Actually, we fix what seems to be a bug inside the theme here. maybe @alorbach can elaborate a bit more on that.

rgerhards:
> @rgerhards See b57c0e4 for how to make the chosen theme preferred (but optional) with a fallback to the default theme should it not be present. If (as with that revision) you omit including the sphinx-better-theme package name in requirements.txt, then (for the most part) systems other than your buildbox (container or not) will generate the docs with the standard/default theme.
> 
> If you're fine with changing the theme for everyone (well, everyone that follows the guidance in the [README](/rsyslog/rsyslog-doc/blob/master/README.md file), then 67cfd32 illustrates the necessary change (ignoring the comments) to the requirements.txt file.

This will work for us, but IMHO a better solution would be to keep the theme selection in a separate file. Than everyone can simply overwrite that file to get what he wants. That a plain dumb solution, and I admit this is why I like it (easy to grasp for everyone). Nevertheless, I'll see that I add your suggestion to current build. If we think a simpler solution is better, we can always change...

deoren:
Well, bundling the theme into the source directory *is* an option. Is that the direction you want to go instead of pulling it in via pip (or using the fallback to default behavior)?

rgerhards:
I think I didn't make myself clear enough. I would like to have a file, e.g. ```theme.inc.conf.py``` with such content inside the git tree:
```
html_theme = 'default'
```
On the web site, I would like to replace it with this:
```
from better import better_theme_path
html_theme_path = [better_theme_path]
html_theme = 'better'
html_theme_options = {'inlinecss': ' @media (max-width: 820px) { div.sphinxsidebar { visibility: hidden; } ',}
```
So I can simply overwrite that file on the server and be all set. And anybody else can do the same.

So no shipment of themes or anything complicated, just an easy way to override the default theme setting. I can do the same with ```sed``` but that's not as reliable as it depends on some fixed values inside ```conf.py``` - and IME sooner or later this breaks. Much less likely with a simple include file.

However, I do not find a way to include this text in ```conf.py```, it looks like python does not have a plain "insert this text from the file here" type of directive. I hoped I just overlook it.

Does this clarify?

source/conf.py
alorbach: physical resolution doesn't match browser resolution when browsing with mobile devices (unless you force desktop mode in chrome on android). So even if the mobile devices would have 4k resolutions, the design would be broken. 

Anyway the theme is kind of broken once the design breaks down to tablet / phone mode. fonts are layed over another, we workaround that problem by hidding the sphinxsidebar. 

To fix this we would have to fix the theme. And then we can also use the current theme (which has little bit better design in my opinion) and make it responsive.

rgerhards: I think currently the theme renders unreadable on small screens. The new theme renders 90% OK. Not great, but much better readable. Nobody I know has time to write an optimized theme. My conclusion is that we should change as proposed and life with the fact that it is an incremental improvement, albeit not a perfect solution.

deoren:
@rgerhards major outage here at work, will respond later when I can.

source/conf.py
deoren: @rgerhards @alorbach Going to respond in detail within the "main flow" of this issue instead of inline here.

deoren:
> I think I didn't make myself clear enough. I would like to have a file, e.g. theme.inc.conf.py with such content inside the git tree:
> 
> `html_theme = 'default'`
>
> On the web site, I would like to replace it with this:
>``` 
> from better import better_theme_path
> html_theme_path = [better_theme_path]
> html_theme = 'better'
> html_theme_options = {'inlinecss': ' @media (max-width: 820px) { div.sphinxsidebar { visibility: hidden; } ',}
> ```
> So I can simply overwrite that file on the server and be all set. And anybody else can do the same.

I see what you mean now, thanks for explaining.

> So no shipment of themes or anything complicated, just an easy way to override the default theme setting. I can do the same with sed but that's not as reliable as it depends on some fixed values inside conf.py - and IME sooner or later this breaks. Much less likely with a simple include file.

I see what you're doing, but from what I understand Sphinx is designed to use themes in one of two forms. From the [doc](http://www.sphinx-doc.org/en/stable/theming.html):

> If the theme does not come with Sphinx, it can be in two static forms: either a directory (containing theme.conf and other needed files), or a zip file with the same contents. Either of them must be put where Sphinx can find it; for this there is the config value html_theme_path. It gives a list of directories, relative to the directory containing conf.py, that can contain theme directories or zip files.

If this is the route that you decide to go, then a subdirectory would be created and the appropriate import statement used (or probably not if `html_theme_path` is set).

> However, I do not find a way to include this text in conf.py, it looks like python does not have a plain "insert this text from the file here" type of directive. I hoped I just overlook it.

Since the theme file is Python code, we would import it and pull in all items into the current namespace. Not sure if there are conflicts with doing that or not. If possible, I suggest we instead treat the file you referenced as part of a theme instead. The folder could be named in such a way that it becomes easy to tell that the theme for the rsyslog-doc project is within and we could update our docs to cover that.

> Does this clarify?

Yes, thanks.

deoren:
@alorbach @rgerhards 

Circling back to my earlier remarks re how the proposed theme displays:

For what it is worth, Mozilla Firefox on a my Windows 10 desktop install has the same problem with this page:

http://www.rsyslog.com/doc/master/configuration/modules/imuxsock.html

In the past I've modified the "Scale and layout" settings from 100% to 125%, but I just double-checked and confirmed that I still have the default 1920x1080 @ 100% scale setting in place.

I tested from an Ubuntu 17.10 VM and replicated the effect, though it wasn't as bad. It may just be that if the decision is made to go with the "better" theme that we need to tweak the styles used by the module docs to work that out.

> @alorbach: physical resolution doesn't match browser resolution when browsing with mobile devices (unless you force desktop mode in chrome on android). So even if the mobile devices would have 4k resolutions, the design would be broken.
> 
> Anyway the theme is kind of broken once the design breaks down to tablet / phone mode. fonts are layed over another, we workaround that problem by hidding the sphinxsidebar.
>
> To fix this we would have to fix the theme. And then we can also use the current theme (which has little bit better design in my opinion) and make it responsive.

Thank you for the feedback. I took a look at the [Sphinx doc for theming](http://www.sphinx-doc.org/en/stable/theming.html) and I'm sure that I found what you guys already have: very little support for customizing the style using theme-provided options.

Based on what you guys provided in this mock-up I explored the styling for the classic theme using the Developer Tools. I believe I found the applicable element and have modified the `rsyslog.css` file to apply the changes to affect at least a little more mobile friendliness. Please see my notes below.

> @rgerhards: I think currently the theme renders unreadable on small screens. The new theme renders 90% OK. Not great, but much better readable. Nobody I know has time to write an optimized theme. My conclusion is that we should change as proposed and life with the fact that it is an incremental improvement, albeit not a perfect solution.

I played with the generated HTML content from this branch a little more, dragging the window to a single column view and stretching it out. I was able to see the effect you were going for and realized why I saw a different (responsive) effect for the Nexus 7 display vs the LG V20 display. 

Still, looking at the changes you've provided thus for got me thinking that if there were design elements you liked from the existing default theme, perhaps we could just modify it slightly to provide the same mobile-friendly functionality.

I took a stab at providing this via the 'responsive-classic-theme-hack' branch on my fork. Here is the output:

http://rsyslog.whyaskwhy.org/responsive-classic-theme-hack/configuration/modules/imuxsock.html

Dragging the browser window about results in some responsive results, but viewing the page on an actual mobile device still results in the sidebar displaying. Perhaps once my test branch is uploaded to the real site we can see what the effect would be then.

In the end, if you opt to replace the theme I will assist with integrating support for the `better` theme as previously demoed or as a local theme directly within the source tree as you feel is needed/appropriate. 

For my part, I believe that requiring the user to read the directions and install required build dependencies (via `pip install -r requirements.txt`) in order to receive the new theme is reasonable, especially if we include the fallback functionality I prototyped in my `rgerhards-site-theme` branch. This results in their not having to do anything and since you'll have the `better` theme installed on your build system then the actual site content will be as you intend. Because CI and your Docker image will reference the `requirements.txt` file you'll still get the intended effect there also.

deoren:
After some additional testing, I have a better appreciation for the mobile support in the `better` theme vs the work to update the default theme to be more mobile friendly.

Ignore my remarks re this prototype branch I put together:

http://rsyslog.whyaskwhy.org/responsive-classic-theme-hack/configuration/modules/imuxsock.html

everything else should still apply.

rgerhards:
> If this is the route that you decide to go, then a subdirectory would be created and the appropriate import statement used (or probably not if html_theme_path is set).

actually I think I did still not make myself clear. I just want to avoid doing
```
sed -i /html_theme = 'default'/html_theme = 'better'/ conf.py
```
and would prefer to "include/import/replace/ a text file instead -- because the sed is not really reliable. 

If ```conf.py``` were C code, I would just say
```
... conf.py settings ...
#include "theme-conf.py"
... more conf.py settings ...
```

Nevertheless, I think you probably have already provided a better solution. I would still be interested to learn if there is a kind of simple "include this text from a file to right here in this file" (like C #include) in python.

deoren:
@rgerhards I forget to mention that I looked closer at the `better` theme and confirmed that it is composed of the main file that you mentioned, but also template files that are processed when generating the output files.

rgerhards:
> For my part, I believe that requiring the user to read the directions and install required build dependencies (via pip install -r requirements.txt) in order to receive the new theme is reasonable, especially if we include the fallback functionality I prototyped in my rgerhards-site-theme branch. This results in their not having to do anything and since you'll have the better theme installed on your build system then the actual site content will be as you intend. Because CI and your Docker image will reference the requirements.txt file you'll still get the intended effect there also.

maybe that's the source of the misunderstanding: my point is that I actually do not even want to force the user to use the rsyslog.com site style. I originally wanted to ship with a default theme and have everyone (including us at rsyslog.com) use the style they want. I was just looking for an easy way to make the required changes automatically. 

Again, I think this is of purely "make sure we understand each other" value -- what you demoed looks good and I'll give the merge a quick look. Still I think we should consider leaving the user the choice of themes and not requiring anything. But as I understand, we ship the current rsyslog theme anyhow, so we already mixing this up for quite a while ;-)

rgerhards:
> After some additional testing, I have a better appreciation for the mobile support in the better theme vs the work to update the default theme to be more mobile friendly.
> 
> Ignore my remarks re this prototype branch I put together:
> 
> http://rsyslog.whyaskwhy.org/responsive-classic-theme-hack/configuration/modules/imuxsock.html
> 
> everything else should still apply.

sorry, want to make sure I understand you correctly. You say that the prototype has not worked out and we should fall back to what is in this PR? I would prefer to nail this soon, at least to a workable state.

If we fall back to (kind of) the PR, I'll probably try to address it via ```sed``` albeit I do not really like it.

deoren:
> @rherhards: I would still be interested to learn if there is a kind of simple "include this text from a file to right here in this file" (like C #include) in python.

AFAIK, that is the import statement. The `from X import *` form pulls all content in from the module and places it into the current namespace. That is probably a close equivalent to what you use with include statements.

There are options to read in content from a text file, but the content is usually then treated as data assigned to one or more variables.

rgerhards:
ok, it looks like python does not support textual include: https://stackoverflow.com/questions/28336318/include-in-c-vs-import-in-python

deoren:
@rgerhards I meant to use branch names for clarity, but re abandoning the work, I was referring to the most recent branch: 7f143a0 or "responsive-classic-theme-hack". The earlier two commits, b57c0e4 and 67cfd32 should still be valid.

rgerhards:
@deoren ok, thx -- but'll follow the ```sed``` path, I really do not want to enforce this on the user. I'll post when I am done, I guess than it becomes clear (sorry for being unable to describe it in words).

rgerhards:
OK, finally got it. I have changed the server script to:
```
echo "======starting doc generation======="
docker pull rsyslog/rsyslog_doc_gen # get fresh version
export DOC_HOME=...
cd $DOC_HOME

for branch in v5-stable v7-stable v8-stable master;
do
        echo "**** $branch ****"
        rm -rf build
        git reset --hard
        git checkout -f $branch
        git pull

        # change config to use rsyslog site theme
        sed -i "s/^[ ]*html_theme = '.*'/from better import better_theme_path\\n\
html_theme_path = \[better_theme_path\]\\n\
html_theme = 'better'/" source/conf.py

        sed -i "s/^.*html_theme_options = {.*}/html_theme_options = {\
        'inlinecss': ' @media (max-width: 820px) { div.sphinxsidebar { visibility: hidden; } }',\
        }/" source/conf.py

        docker run -ti --rm \
                -u `id -u`:`id -g` \
                -e STRICT="" \
                -e SPHINX_EXTRA_OPTS="-q" \
                -v "$DOC_HOME":/rsyslog-doc \
                rsyslog/rsyslog_doc_gen
        rsync -rz build/* ...
done
```
No changes are required on the project itself. So this separates site appearance from whatever default theme is. I hope this clarifies the original intent.

rgerhards:
automatic builds are now activated again

deoren:
> @rgerhards: ok, it looks like python does not support textual include:
>  https://stackoverflow.com/questions/28336318/include-in-c-vs-import-in-python

That is true, and is in a way what I was saying earlier regarding the import statement. Your goal seemed to be to pull in a file and have it inserted at the place the import occurred. This can be emulated with an import statement that drops all content into the current namespace. It's not as straightforward, but it can be done. On a related note though, I went back and learned that the `better` theme is composed of many files, the conf file you referenced earlier, two CSS *template* files that are processed at build time and some HTML templates as well.

Regarding your earlier comments of me not understanding what you were saying, I agree that we have probably been talking past each other. I think one thing that may have not helped is the shortage of time that you have to spend on tangent goals like this one and my overly verbose replies that I made in an attempt to address all items that you covered. Those two items combined make dealing with long-running tasks (that you intended to be really simple and quick) more aggravating and draining than they should be. I probably need to work on focusing my responses so that they're easier to follow, and I will need to start splitting off into separate PRs any attempts on my part to showcase alternatives to what you're doing.

> @rgerhards:  ok, thx -- but'll follow the sed path, I really do not want to enforce this on the user

That is actually one of the points I did not get across. With the changes I was proposing in b57c0e4 and 67cfd32, the user did not have to do anything extra. They just built the docs and ended up with either generated docs using the default theme or, if they happened to have the `better` theme installed, that theme would be used instead. 

For your case, the theme would be present, either because you installed it manually or because your setup process would call `pip install -r requirements.txt`. As I said before, I recommend that you do not embed the theme within the repo and I also recommend that you do not fall back to using `sed` to wire up the configuration as part of your Docker build script. The biggest reason why is that it becomes unclear to your doc team (particularly new members coming on in the future) what theme you intend to use. If you go with b57c0e4 and 67cfd32, it comes very clear what the intent is.

I will submit a new PR that includes the branch whose commits I just referenced so that we can (briefly) discuss this further there.

rgerhards:
On the phone: I think we still do not talk along the same lines ;-) but maybe I found the/one source of confusion: the conf file I was talking about is ./source/conf.py - I am not aware of any other one.

Secondly I am not sure if that issue would better be solved by live (voice) conversation.

deoren:
@rgerhards Earlier in the discussion you mentioned theme.conf.in.py (https://github.com/rsyslog/rsyslog-doc/pull/533#issuecomment-362525318). I was sharing that unfortunately (or not, depending on how you look at it), the Sphinx themes are composed of many files and it's not feasible to collapse them into one theme file that you drop into the repo.

Of course as I say that, I think I see what you were getting at. Perhaps the pieces are falling into place for me now; you were not speaking of including the theme itself (as I thought you were), but were perhaps speaking of an include file that specified *using* the theme. If that is what you meant, there are ways we could do that, including importing a separate configuration file. If the file is present its settings are used, if not then the settings already in `source/conf.py` are used. 

I still think PR #536 is a workable solution, at least for right now. That PR attempts similar logic, but with including and then using the `better` theme.

I originally thought you were looking to include the theme file itself, which in this case isn't one file, but many (theme conf file, Jinja2 theme CSS and HTML templates that are processed at build time):

https://github.com/irskep/sphinx-better-theme/tree/master/better

rgerhards:
Need to reply tomorrow fully, but the script I posted really does all I need. I just wanted to put the two settings in a separate file, but sed also works. As I said, will elaborate tomorrow. Sorry for the confusion caused by the name I used. Sounded innocent to me...

deoren:
> @rgerhards: Need to reply tomorrow fully, but the script I posted really does all I need. I just wanted to put the two settings in a separate file, but sed also works. As I said, will elaborate tomorrow. 

> @rgerhards: Sorry for the confusion caused by the name I used. Sounded innocent to me...

Not a problem, and it's just as likely that the confusion is all internal and had nothing to do with your wording. Sorry if I came off as waspish in my responses. If I did, it wasn't intentional.

If you're up for it, let's move the discussion on this over to PR #536?

> @rgerhards: Secondly I am not sure if that issue would better be solved by live (voice) conversation.

Ha, if your ears can stand my southern US accent I'm game for trying to interpret your German accent. I suppose the stereotype of Americans struggling to understand other accents probably applies to me. ;)

Tangent: I'm in a UTC-6 timezone, so I think that places us about 7 hours apart. Considering how often we've conversed later in the evening here I'd guess you're having to stay up late to respond to me. I'm fine with waiting on a response until the next day; just because I have poor sleep habits doesn't mean I have to corrupt others with them. :P

Issue #535, opened 2018-02-03
v8-stable still uses temporary version number


rgerhards:
@deoren I experimentally merged master into a copy of v8-stable and rebuild it. However, the version number was still the temporary one. I assumed that the stable build read the version number from a file. I just had a quick glimpse, but that was not the case.

If I am right, I suggest we

* add a file VERSION and RELEASE, which match the previos ```version``` and ```release``` variables
* if not present, use the current temporary version ID algo
* if present, use the values from there

That we we need to update these files only on release and can be sure that whenever someone builds from v8-stable the version numbers are pinned.

As a tangent, I would like to shorten the temporary version number used at the rsyslog.com site, it requires a lot of space. Do you have any concerns in this regard? If so, let's open a new tracker for the discussion.

deoren:
@rgerhards The current configuration is using a similar setup to what you've described here. I know you've got a lot going on, but please carve out some time to read over the first part of this doc: 

https://github.com/rsyslog/rsyslog-doc/blob/master/BUILDS_README.md

and my summary of the changes here:

https://github.com/rsyslog/rsyslog-doc/pull/493#issuecomment-359163281

Because you showed a preference for using the `release_build.sh` script to generate the docs and release tarball at release time, that script was updated to provide the required results.

> @rgerhards: As a tangent, I would like to shorten the temporary version number used at the rsyslog.com site, it requires a lot of space. Do you have any concerns in this regard? If so, let's open a new tracker for the discussion.

As you noted, let's discuss on a separate issue. I'll create one so that we can cover that topic there.

rgerhards:
Ok I need to re-read it. That means I need to change the doc update process so that I now find the latest tarball (I think). Right now I just do a gut checkout and be done.

rgerhards:
I posted the current update script in the other tracker.

deoren:
> @rgerhards: That means I need to change the doc update process so that I now find the latest tarball (I think). Right now I just do a gut checkout and be done.

I think it would help to illustrate. 

Starting at line 66 and on through line 70 is where the release build number logic is used to pull the version and release string details that are used later in the substitution process:

https://github.com/rsyslog/rsyslog-doc/blob/a66e62366d213d1cde4f647ff087463fe8730421/release_build.sh#L66


Starting here and continuing until line 118 is where the placeholder values are substituted with the latest stable tag values:

[...]

PR #536, opened 2018-02-03
Attempt to use 'better' theme, fall back to default theme if not available

deoren:
This PR is closely related to the discussion from rsyslog/rsyslog-doc#533.

There are two things accomplished with these changes:

1. Explicitly specify in `requirements.txt` that this Sphinx documentation project relies upon (at present), two components:
    1. Sphinx
    1. `better` theme

1. Attempt to *use* the `better` theme, but if not present gracefully fall back (without error messages) to using the default theme used previously.

This PR does not touch the current Dockerfile, but after this PR is merged it could be updated to take advantage of the changes introduced here (such as using `pip install -r requirements.txt` instead of hard-coding specific packages on a `RUN` line).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,9 @@
 
 # Need at least 1.5.1 due to some of the features we are using
 sphinx >=1.5.1
+
+# Preferred theme; integrates well with the official website
+# Note: the docs will build without this theme package present,
+# but it is preferred that it be used over the stock 'default'
+# theme due to the enhancements regarding whitespace
+sphinx-better-theme
--- a/source/conf.py
+++ b/source/conf.py
@@ -14,6 +14,19 @@
 import os
 import sys
 
+
+# Attempt to import our desired theme. If successful, then we will use that
+# theme to build the docs. If the import attempt fails, we will use the
+# default theme instead.
+try:
+    from better import better_theme_path
+    html_theme_path = [better_theme_path]
+    desired_theme_available = True
+
+except ImportError:
+    desired_theme_available = False
+
+
 # Module for our custom functions. Used with automating automating build info.
 sys.path.append(os.getcwd())
 import conf_helpers
@@ -212,14 +225,35 @@ suppress_warnings = ['epub.unknown_project_files']
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
+
+# The theme used for building was set earlier based on the results of an
+# import attempt. If the import was successful, our desired theme will be
+# used. If the import failed, the 'default' theme will be used instead.
+# This works around the possibility that our users will fail to read
+# the direction to run `pip install -r requirements.txt` and will instead
+# just install the sphinx package directly (either via pip install sphinx
+# or via their distro packaging system).
+if desired_theme_available:
+    html_theme = 'better'
+else:
+    html_theme = 'default'
+
+#html_theme = 'default'
 #html_theme = 'basic'
 #html_style = 'rsyslog.css'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-#html_theme_options = {}
+if html_theme == 'better':
+    # Enable max width for our chosen theme only
+    html_theme_options = {
+    'inlinecss': ' @media (max-width: 820px) { div.sphinxsidebar { visibility: hidden; } }',
+    }
+
+else:
+    # If our chosen theme is not present, reset the options dictionary
+    html_theme_options = {}
 
 # Add any paths that contain custom themes here, relative to this directory.
 #html_theme_path = []


deoren:
The more I work with this the more I think that the sed calls can be replaced with `-D` sphinx-build overrides and the `source/conf.py` changes I made can be dropped entirely.

I'm on the fence regarding adding the `better` theme to the `requirements.txt` file. On one hand it's easy enough to include it in there, but unless the user opts to actually use it then nothing is really gained. In that case, I'd recommend two things:

1. Adding the `better` theme to `requirements.txt`
1. Changing the default within `source/conf.py`

This would be *mostly* safe since other projects would override the default theme anyway with their own build override settings. The one catch might be the `inlinecss` theme option specified for the `better` which isn't supported by the classic theme.

I guess in the end the real fix is to just override the default using `-D` sphinx-build options within your build script. This avoids using sed calls that have the potential to break when text is shifted about within `source/conf.py` and you *know* which settings you want to have applied.

We can always explore this further if you opt to change the default theme in the future. Perhaps at that point we could bundle the combination of theme choices/modifications into a fork that we maintain. That fork could then be referenced in the `requirements.txt` file (perhaps even directly from GitHub).

rgerhards:
> I guess in the end the real fix is to just override the default using -D sphinx-build options within your build script. This avoids using sed calls that have the potential to break when text is shifted about within source/conf.py and you know which settings you want to have applied.

Can I just use ```-D html_theme='better'``` and this is not overridden by conf.py? If so, and if that works for the other settings as well, that was actually what I was looking for since mid last week ;-) I just didn't know that this would be possible.

deoren:
> @rgerhards: Can I just use -D html_theme='better'

It will depend on the theme name. Some themes I've seen list additional steps, but that may be for older versions of Sphinx. I'll test and report back.

> @rgerhards: and this is not overridden by conf.py?

The `-D` command-line setting wins out over anything in the `conf.py` file, which is how I believe that readthedocs.org and others are able to easily deal with the defaults already defined in the `conf.py` file in this project. I need to check the Docker image for readthedocs.org to see how they've got their build process setup.

> @rgerhards:  If so, and if that works for the other settings as well, that was actually what I was looking for since mid last week ;-) I just didn't know that this would be possible.

It's probably something we both would have seen had one of us been looking over the other's shoulder. Oh well, better late than never. :)

deoren:
This is an example of specifying the theme via command-line when I know what my virtualenv is:

`sphinx-build -D html_theme="better" html_theme_path="$HOME/rsyslog-docs-build/lib/python2.7/site-packages" -b html source build`

and if the package is installed site-wide it looks like this:

`sphinx-build -D html_theme="better" html_theme_path="/usr/local/lib/python2.7/dist-packages" -b html source build`

rgerhards:
> It's probably something we both would have seen had one of us been looking over the other's shoulder. Oh well, better late than never. :)

IME this is really one of the areas where interactive (voice) calls work best. I bet it would have taken no more than 10 minutes to sort it out. Unfortunately it's pretty inconvenient due to TZ differences. But maybe it would make sense to schedule a Google Hangout on Air every now and then. So whoever wants can participate and we could also record the session to be transparent for others. Just an idea, and an inconvenient one (at least for me). ;-)

rgerhards:
@deoren Can you have a closer look some time this week at what I do with ```sed```. I fear the "better" them requires some more setup, but maybe it's still doable with ```-D```. That info, BTW, solves many of my concerns.

deoren:
> @rgerhards: @deoren Can you have a closer look some time this week at what I do with sed. 

I did. From what I can see you're performing this substitution:

```diff
-html_theme = 'default'
+from better import better_theme_path
+html_theme_path = [better_theme_path]
+html_theme = 'better'
```

where 'default' could be any other theme previously set. I tested against that scenario when I posted my examples. The `-D html_theme_-ath="..."` bit from my examples is what allows it to work. 

I didn't look at the source for the themes, but I'm pretty sure the first two lines in the substitution shown above is just to work around having to explicitly specify the path to either the `site-packages` or `dist-packages` directory.

deoren:
> @rgerhards: IME this is really one of the areas where interactive (voice) calls work best. I bet it would have taken no more than 10 minutes to sort it out. 

Maybe so.

> @rgerhards: Unfortunately it's pretty inconvenient due to TZ differences. 

Are you in UTC+1?

> @rgerhards: But maybe it would make sense to schedule a Google Hangout on Air every now and then. So whoever wants can participate and we could also record the session to be transparent for others. Just an idea, and an inconvenient one (at least for me). ;-)

Worth at shot at some point. I don't know that I've ever used Google Hangout on Air, but it shouldn't be too hard to figure out.

rgerhards:
> @rgerhards: IME this is really one of the areas where interactive (voice) calls work best. I bet it would have taken no more than 10 minutes to sort it out.
> 
> Maybe so.

The key thing for me is to be able to ask a question right away at the right spot ... and notice if "the other one" begins to be puzzled. With writing, you often write loooots of text  ... just to learn that it is not understood or there is a question right at the beginning. That's also a reason I nowadays tend to pretty short responses...

> Are you in UTC+1?

yup... As I tend to always confuse +1/-1 UTC, this page may be useful for you as well:

https://www.timeanddate.com/worldclock/germany/wurzburg

:-)

> Worth at shot at some point. I don't know that I've ever used Google Hangout on Air, but it shouldn't be too hard to figure out.

As a typical Google tool, it changes constantly so prior experience is not always helpful ;-) I noticed that yesterday...

deoren:
> @rgerhards: yup... As I tend to always confuse +1/-1 UTC, this page may be useful for you as well:
> 
> https://www.timeanddate.com/worldclock/germany/wurzburg

Thanks. It was somewhat of a guess on my part based on something I read of yours in the past. I've bookmarked that for future reference.

> @rgerhards: The key thing for me is to be able to ask a question right away at the right spot ... and notice if "the other one" begins to be puzzled. With writing, you often write loooots of text ... just to learn that it is not understood or there is a question right at the beginning. That's also a reason I nowadays tend to pretty short responses...

I hear you, I tend to rely pretty heavily on immediate feedback as well when covering topics. If we run up against sticking points in the future I'm up for attempting a Hangouts conversation. I'll need to dig up a microphone for it and plan ahead, but it should be doable on occasion.

deoren:
> @rgerhards: I fear the "better" theme requires some more setup, but maybe it's still doable with -D

I missed it when I first read your reply, but yeah, looks like there will need to be three overrides passed to `sphinx-build`. It should still be doable though. As part of the discussion on #535 I'm putting together a prototype version of the `release_build.sh` script that will use the reference Docker image to build stable release docs (and tarball).

I'll post as a new PR and reference it from here for further discussion.

deoren:
I've confirmed this works from inside the reference container:

```
sphinx-build -D html_theme="better" -D html_theme_path="/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages" -D html_theme_options.inlinecss="@media (max-width: 820px) { div.sphinxsidebar { visibility: hidden; }" -b html source build
```

The trick is passing that into the container with quoting intact. Head ... wall ...

rgerhards:
excellent! let me see how this integrates with the script.

deoren:
@rgerhards I'm still fighting with it and not having much luck. Any chance a separate "tool" command file can be added for building for rsyslog.com vs general builds? (e.g., bash, git-clone, help, ...)

rgerhards:
There must be a general solution. I am right now looking into it. Right now, I am not sure if it is an issue with Alpine's ash. In theory, this should work flawless. Give me a bit...

deoren:
@rgerhards np. I created PR #543 to demo a workaround for the issue.

rgerhards:
No, it's not ash. It's strange. The quoting seems intact, but somehow the final quotes are not removed. If I send the command line to a file and then execute that file, everything work:
```
/rsyslog-doc $ sphinx-build -b html $SPHINX_EXTRA_OPTS source build
Running Sphinx v1.6.7
loading pickled environment... not yet created

Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/sphinx/cmdline.py", line 305, in main
    opts.warningiserror, opts.tags, opts.verbosity, opts.jobs)
  File "/usr/lib/python2.7/site-packages/sphinx/application.py", line 234, in __init__
    self._init_builder()
  File "/usr/lib/python2.7/site-packages/sphinx/application.py", line 311, in _init_builder
    self.builder.init()
  File "/usr/lib/python2.7/site-packages/sphinx/builders/html.py", line 256, in init
    self.init_templates()
  File "/usr/lib/python2.7/site-packages/sphinx/builders/html.py", line 299, in init_templates
    self.theme = theme_factory.create(themename)
  File "/usr/lib/python2.7/site-packages/sphinx/theming.py", line 285, in create
    '(missing theme.conf?)') % name)
ThemeError: no theme named u'"better"' found (missing theme.conf?)

Theme error:
no theme named u'"better"' found (missing theme.conf?)
/rsyslog-doc $ echo sphinx-build -b html $SPHINX_EXTRA_OPTS source build >tmp
/rsyslog-doc $ source ./tmp
Running Sphinx v1.6.7
loading pickled environment... not yet created
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 254 source files that are out of date
updating environment: 254 added, 0 changed, 0 removed
reading sources... [  0%] community
reading sources... [  0%] compatibility/index
```

rgerhards:
```
ash -c "sphinx-build -b html $SPHINX_EXTRA_OPTS source build"
```
also works... maybe I am overlooking some fine detail in sh command line expansion...

deoren:
> @rgerhards: If I send the command line to a file and then execute that file, everything work

It's not ideal, but it could work. What do you think about a quick workaround of using a separate "command" file?

https://github.com/deoren/rsyslog-doc/blob/7691c4b5f1f45bb4a5adb6850ae8b724905c4304/tools/buildenv/tools/build-website-doc#L22

rgerhards:
I really don't like this. It's not clean and if we use it, it will probably stay for years. Plus we'll add more such exceptions. The ```ash -c ``` prefix is much cleaner (generic). I also think there is a better solutions. Let's nail it and not give up too early ;-)

deoren:
> @rgerhards: I really don't like this. It's not clean and if we use it, it will probably stay for years.

Are you referring to the workaround of echoing out to a file and then executing it or creating a new "tools" entry?

rgerhards:
This: https://github.com/rsyslog/rsyslog-doc/pull/536#issuecomment-363344872

rgerhards:
I misunderstood. I don't like adding a tool just for the website.

rgerhards:
It's a sh ```eval``` vs not issue. I guess my understanding of the shell is wrong here. That needs to be fixed. I'll craft a small test program to see what the command line params actually look.

Also, this brought up that passing complex parameters is a pain. We should provide an ability to pass them in via a file. For example, if I have a file ```sphinx_extra_options``` which contains
```
-v
-D html_theme_path="/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages"
-D html_theme=better
```
and I do
```
eval "sphinx-build  -b html `tr '\n' ' ' < sphinx_extra_opts`  source build"
```
that works.

Now questions are:
* what is the actual root cause
* does it make sense to support such an optional file (I think so)

To explain a bit more on the extra tools file. This is what I call "giving up" in software engineering. It's often appropriate, but the art of software engineering is to keep abstraction layers. I usually tend to be a bit persistent before I write non-generic stuff. In the longer term that pays out. Otherwise you'll end of with many copies for many different cases -- which you'll screw up to consistently update at a time. This thinking is probably over-engineering for a script like this, but I treat the **container** as a software development project. If it just were a cron script, I would have hesitated a moment to apply the workaround. But it's more IMHO (again, maybe overdoing).

deoren:
> `sphinx-build -b html $SPHINX_EXTRA_OPTS source build`

What about `sphinx-build -b html "$SPHINX_EXTRA_OPTS" source build` ?

Here is the effect of splitting out the settings to separate variables and then referencing them on the command-line:

```bash
sphinx_html_theme="better"
sphinx_html_theme_path="/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages"
sphinx_html_theme_options="@media (max-width: 820px) { div.sphinxsidebar { visibility: hidden; } }"

sphinx-build -D html_theme="$sphinx_html_theme" -D html_theme_path="$sphinx_html_theme_path" -D html_theme_options.inlinecss="$sphinx_html_theme_options" -b html source build
```

Output:

```
/rsyslog-doc # sphinx-build -D html_theme="$sphinx_html_theme" -D html_theme_path="$sphinx_html_theme_path" -D html_theme_options.inlinecss="$sphinx_html_theme_options" -b html source build
Running Sphinx v1.6.6
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 250 source files that are out of date
updating environment: [config changed] 254 added, 10 changed, 0 removed
^Cading sources... [  3%] compatibility/v8compatibility                                                                                                                                                            
interrupted!
```

rgerhards:
> One reason I leaned in that direction aside from applying a workaround until a better solution was found was because the container was already setup to process/build for the website (the better theme is package inside of it for example).

Yeah, that's actually a good sample of what I mean. I screwed up when I added the better theme. With it, it gives reason to more such changes, which gives reason to... I really should correct that mistake when done with the current issue.

rgerhards:
> sphinx-build -b html $SPHINX_EXTRA_OPTS source build
> 
> What about sphinx-build -b html "$SPHINX_EXTRA_OPTS" source build ?
> 
> Here is the effect of splitting out the settings to separate variables and then referencing them on the command-line:
> 
> sphinx_html_theme="better"
> sphinx_html_theme_path="/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages"
> sphinx_html_theme_options="@media (max-width: 820px) { div.sphinxsidebar { visibility: hidden; } }"
> 
> sphinx-build -D html_theme="$sphinx_html_theme" -D html_theme_path="$sphinx_html_theme_path" -D html_theme_options.inlinecss="$sphinx_html_theme_options" -b html source build
> Output:
> 
> /rsyslog-doc # sphinx-build -D html_theme="$sphinx_html_theme" -D html_theme_path="$sphinx_html_theme_path" -D html_theme_options.inlinecss="$sphinx_html_theme_options" -b html source build
> Running Sphinx v1.6.6
> loading pickled environment... done
> building [mo]: targets for 0 po files that are out of date
> building [html]: targets for 250 source files that are out of date
> updating environment: [config changed] 254 added, 10 changed, 0 removed
> ^Cading sources... [  3%] compatibility/v8compatibility                                                                                                                                                            
> interrupted!

bear a bit with me... I like to reproduce it. Right now, I am really interested in the root cause, not just in order to solve the issue...

deoren:
> @rgerhards: bear a bit with me... I like to reproduce it. Right now, I am really interested in the root cause, not just in order to solve the issue...

Sure. Was just tossing ideas out to see if something "clicked" when you saw what I shared. I dealt with this years ago, but I can't recall too much about what the actual fix was.

rgerhards:
ok, same problem reproduces outside of the container in an debian box:
```
rger@debian:~/rsyslog-doc$ export OPT='-D html_theme_path="/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages" -D html_theme="better"'             
rger@debian:~/rsyslog-doc$ echo $OPT 
-D html_theme_path="/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages" -D html_theme="better"
rger@debian:~/rsyslog-doc$ echo $OPTS^C
rger@debian:~/rsyslog-doc$  sphinx-build  -b html $OPT source buildRunning Sphinx v1.6.5
loading pickled environment... not yet created

Theme error:
no theme named u'"better"' found (missing theme.conf?)

```

rgerhards:
aha!

small test program ```argv.c```:
```
#include <stdio.h>

int main(int argc, char *argv[])
{
	int i;
	printf("argc: %d\n", argc);
	for(i = 0 ; i < argc ; ++i)
		printf("argc[%d]: %s\n", i, argv[i]);
}
```
Then:
```
rger@debian:~/rsyslog-doc$ ./argv  -b html $OPT source build
argc: 9
argc[0]: ./argv
argc[1]: -b
argc[2]: html
argc[3]: -D
argc[4]: html_theme_path="/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages"
argc[5]: -D
argc[6]: html_theme="better"
argc[7]: source
argc[8]: build
rger@debian:~/rsyslog-doc$ eval ./argv  -b html $OPT source build
argc: 9
argc[0]: ./argv
argc[1]: -b
argc[2]: html
argc[3]: -D
argc[4]: html_theme_path=/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages
argc[5]: -D
argc[6]: html_theme=better
argc[7]: source
argc[8]: build
```
Note the difference between the eval and the non-eval!
The try with parameters directly on command line (instead of passed in via one env var):
```
rger@debian:~/rsyslog-doc$ ./argv  -b html -D html_theme_path="/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages" -D html_theme="better" source build
argc: 9
argc[0]: ./argv
argc[1]: -b
argc[2]: html
argc[3]: -D
argc[4]: html_theme_path=/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages
argc[5]: -D
argc[6]: html_theme=better
argc[7]: source
argc[8]: build
```
Of course (lol), it is the shell that removes the quote chars, but obviously (now) not if they are passed in via a single environment var.

rgerhards:
If I remove the quotes inside the env var, all also works:
```
rger@debian:~/rsyslog-doc$ export OPT='-D html_theme_path=/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages -D html_theme=better'
rger@debian:~/rsyslog-doc$ ./argv  -b html $OPT source build
argc: 9
argc[0]: ./argv
argc[1]: -b
argc[2]: html
argc[3]: -D
argc[4]: html_theme_path=/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages
argc[5]: -D
argc[6]: html_theme=better
argc[7]: source
argc[8]: build
```
I also tried with sphinx. But now we have some param with spaces. I did ignore that for now. Let's see...

rgerhards:
... and that does not work out (as expected):
```
rger@debian:~/rsyslog-doc$ export OPT='-D html_theme=better -D html_theme_path="/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages" -D html_theme_options.inlinecss="@media (max-width: 820px) { div.sphinxsidebar { visibility: hidden; }"'rger@debian:~/rsyslog-doc$ !.
./argv  -b html $OPT source build
argc: 19
argc[0]: ./argv
argc[1]: -b
argc[2]: html
argc[3]: -D
argc[4]: html_theme=better
argc[5]: -D
argc[6]: html_theme_path="/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages"
argc[7]: -D
argc[8]: html_theme_options.inlinecss="@media
argc[9]: (max-width:
argc[10]: 820px)
argc[11]: {
argc[12]: div.sphinxsidebar
argc[13]: {
argc[14]: visibility:
argc[15]: hidden;
argc[16]: }"
argc[17]: source
argc[18]: build
rger@debian:~/rsyslog-doc$ sphinx-build  -b html $OPT source build
Error: Cannot find source directory `/home/rger/rsyslog-doc/(max-width:'.
```
But it now makes a lot of sense that ```eval``` works - because it actually does the required re-evaluation of the command line, reparsing the parameters. Let me think about it, but I guess we have just hit a case where ```eval``` is actually needed and the right thing to do...

rgerhards:
I should have added:
```
rger@debian:~/rsyslog-doc$ eval ./argv  -b html $OPT source build
argc: 11
argc[0]: ./argv
argc[1]: -b
argc[2]: html
argc[3]: -D
argc[4]: html_theme=better
argc[5]: -D
argc[6]: html_theme_path=/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages
argc[7]: -D
argc[8]: html_theme_options.inlinecss=@media (max-width: 820px) { div.sphinxsidebar { visibility: hidden; }
argc[9]: source
argc[10]: build
```

rgerhards:
There are a couple of security concerns around eval, e.g. here http://mywiki.wooledge.org/BashFAQ/048

I think they do not apply much to our use case, but still some need to think...

deoren:
> There are a couple of security concerns around eval, e.g. here http://mywiki.wooledge.org/BashFAQ/048
> 
> I think they do not apply much to our use case, but still some need to think...

I recall that bit. I'll have to try and remember where I ran into the issue where it appeared I required eval. If I remember correctly, I was dynamically building a command for execution, much like it appears we're doing here. That is where I read the concerns about doing so (which makes sense), but I think my use case wasn't where user input was accepted.

Memory is too foggy on that point. Speaking of which, I'll have to come back to this later I think (need sleep).

Good luck with this.

rgerhards:
have a great sleep (or had it ;-)). I thought about the security issue, but it is none in our case:

* for one, we are exactly trying to give the user great power
* also, there are ample opportunities to run any command the user wishes in his own context, so there really is no extra risk added by the eval

conclusion: eval is the right fix.

We should also think if a file for adding optional settings make sense. I think so, it's way easier than fighting with the multi-invocation command line. Will add a tracker item.

rgerhards:
With https://github.com/rsyslog/rsyslog-doc/pull/545 applied, this script now works:
```
# to turn on container debug mode, add this to the docker command below:
# -e DEBUG="on"
echo "======starting doc generation======="
docker pull rsyslog/rsyslog_doc_gen # get fresh version
cd $DOC_HOME

for branch in v5-stable v7-stable v8-stable master;
do
        echo "**** $branch ****"
        rm -rf build
        git reset --hard
        git checkout -f $branch
        git pull

        docker run -ti --rm \
                -u `id -u`:`id -g` \
                -e STRICT="" \
                -v "$DOC_HOME":/rsyslog-doc \
                -e SPHINX_EXTRA_OPTS='-q -D html_theme_path="/usr/lib/python2.7/site-packages,/usr/local/python2.7/dist-packages" -D html_theme="better"  -D html_theme_options.inlinecss="@media (max-width: 820px) { div.sphinxsidebar { visibility: hidden; }"' \
                rsyslog/rsyslog_doc_gen
        rsync ...
done
```

rgerhards:
I now uploaded the script as well to the rsyslog-website project:

https://github.com/rsyslog/rsyslog-website/blob/master/tools/gendoc.sh

As you can see, the credential issue was easy to solve... It's not yet active, I have to admit. Hope to have it soon.

deoren:
> I now uploaded the script as well to the rsyslog-website project:
> 
> https://github.com/rsyslog/rsyslog-website/blob/master/tools/gendoc.sh
> 
> As you can see, the credential issue was easy to solve... It's not yet active, I have to admit. Hope to have it soon.

ACK. I will try to update the other tickets later tonight.

deoren:
> @rgerhards: I now uploaded the script as well to the rsyslog-website project:
>
> 
> https://github.com/rsyslog/rsyslog-website/blob/master/tools/gendoc.sh
> 
> As you can see, the credential issue was easy to solve... It's not yet active, I have to admit. Hope to have it soon.

That's pretty clever. I've used environment variables set within cron jobs before as an alternative to local include/config files, but your extensive use of that mechanic is causing me to rethink my habits. If nothing else, even if I don't use that approach I'll at least consider it when looking at my options.

PR #540, opened 2018-02-05
Build: Set 'simple' dev build string as default

deoren:
## Previous default build string

Prior hard-coded dev build string:

  `rsyslog 8.xx-BRANCH-YYYYMMDD-commit_hash documentation`

That format has been given a name of 'detailed'. 

## New default build string

A new simpler format has been defined as:

  `rsyslog commit_hash docs`

## Stable releases

Stable release build strings still use the format of:

  `rsyslog 8.x.x documentation`

## epub format

As before, epub release builds default to the release build strings. epub dev builds still use the detailed release string format. This is because the display of epub files is not affected in the same way as HTML output using themes that are more narrowly restricted in horizontal space.

## How to test

*Assuming you already have Python and Sphinx installed ...*

1. git clone https://github.com/deoren/rsyslog-doc rsyslog-doc-testing
1. cd rsyslog-doc-testing
1. git checkout i537-short-dev-build-release-string
1. sphinx-build -b html source build
1. sphinx-build -b epub source build
1. firefox build/index.html

## Example output

Live view:

- http://rsyslog.whyaskwhy.org/i537-short-dev-build-release-string/
- http://rsyslog.whyaskwhy.org/i537-short-dev-build-release-string/rsyslog.epub

I've rebased since that build completed (fixed some doc comments), so the commit hashes won't match up if you compare them. When the next build completes they should match up then.

closes rsyslog/rsyslog#537

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -125,6 +125,17 @@ if version == '8':
         # fill in values that were not specified by a build script.
         release_type = 'dev'
 
+        # Valid options are "simple" and "detailed". "Simple" contains a short
+        # dev build string intended for space constrained display. "Detailed"
+        # contains a much longer dev build string intended for output that is
+        # used for for copyediting/proofreading, where the generated docs may
+        # hang around a while before being disposed of. In that situation,
+        # having detailed build details in the doc are useful to help
+        # differentiate multiple document builds (even from different branches)
+        # from each other.
+        #release_string_detail = 'detailed'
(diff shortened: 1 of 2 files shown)


deoren:
> @rgerhards: I cannot tell exactly about the inner Python workings, but the result looks good to me. Go!

Thanks, will merge now.

@davidelang Please sound off if you notice anything amiss and we can revisit this.

PR #543, opened 2018-02-06
release build script: "wrap" reference Docker image

deoren:
Instead of attempting to control sphinx-build directly this script now attempts to use the reference Docker image to build a stable release of the docs.

To work around issues with passing the required values via environment variables a new "tool" entry was created to allow for optional building of the docs using the 'better' theme.

As part of prototyping these changes I rebuilt the Docker image and uploaded to my Docker Hub profile. That is currently hard-coded in the script for demo purposes.

refs #536, #535


rgerhards:
I think all is clear based on the other issue trackers. If not, let me know.

deoren:
A few things to clarify:

- will the `release_build.sh` script be used to generate official release tarballs or do you have another script in mind for that?
- if the script will be used to build the release tarball, will it also generate content for the website?

rgerhards:
> will the release_build.sh script be used to generate official release tarballs or do you have another script in mind for that?

just for the official tarballs

> if the script will be used to build the release tarball, will it also generate content for the website?
no

rgerhards:
> will the release_build.sh script be used to generate official release tarballs or do you have another script in mind for that?
> 
> just for the official tarballs

to be more precise: it shall do what is required for **building a release** -- and nothing that has not to do with it. In the light of other discussions, this may also involve update ```conf.py``` to carry the fixed release version. And maybe some other things that come up during release build. 

But I see a very small and slim script, focussing on supporting the release build process. Anything else IMHO should get it's own script.

deoren:
> @rgerhards:  I would not force to use the container for building the release (but also would not object if we do it). There is of course some logic in using the container, but again the only target audience is rsyslogd-doc maintainers that do releases. 

Based on feedback from mbiebl, we know that he throws out the generated content and rebuilds from source. Presumably other package maintainers do this as well. 

Do you plan to distribute docs generated using the `better` theme or the classic theme? As you noted, the `better` theme is being used in the `rsyslog/rsyslog_doc_gen` docker image. If your goal is to provide docs built with that theme then it sounds like using the docker image is the way to go as this will also automatically update the generated docs in the future to use whatever theme is the preferred theme at that point in time.

If you prefer to continue distributing the classic theme, then omitting use of the docker image for now is probably the way to go.

rgerhards:
> Do you plan to distribute docs generated using the better theme or the classic theme? As you noted, the better theme is being used in the rsyslog/rsyslog_doc_gen docker image. 

That hack was indeed a big mistake I made. I'll fix it soon. It does not belong there.

> If your goal is to provide docs built with that theme then it sounds like using the docker image is the way to go as this will also automatically update the generated docs in the future to use whatever theme is the preferred theme at that point in time.

I do not think we should force any specific theme.

On the ```release_build``` script: after I finally (... ;-)) realized that I can simply overwrite the versions inside ```conf.py``` I consider this issue as solved. I do not think we urgently need to spend time on changing anything. As it looks, I can continue to use the same release process that proved to work very well during the past 5 years or so:

1. merge master into v8-stable
1. update version info
1. create tarball
1. do checks, iterate on issues, if all well tag and be done

IMHO this is sufficient, especially thinking about the big TODO list we have in front of us.

deoren:
> @rgerhards: IMHO this is sufficient, especially thinking about the big TODO list we have in front of us.

Acknowledged. I meant to circle back to this before now, but got busy with other things. Sorry about that.

I'll close this out and toss the branch. If future work is done it will be better accomplished with fresh eyes anyway.

PR #545, opened 2018-02-06
container: fix quoted parameter passing to container build

rgerhards:
PR now contains a bit more, see individual commits. Still think all of these features make sense together, and sparing time to do separate PRs (that would need to build on each other).

### DIFF
--- a/tools/buildenv/tools/build-doc
+++ b/tools/buildenv/tools/build-doc
@@ -14,6 +14,9 @@ else
 	export OUTPUT=build
 fi
 cd /rsyslog-doc
+if [ -e SPHINX_EXTRA_OPTS ]; then
+	export SPHINX_EXTRA_OPTS="$SPHINX_EXTRA_OPTS `cat SPHINX_EXTRA_OPTS`"
+fi
 eval sphinx-build $STRICT -b $FORMAT $SPHINX_EXTRA_OPTS source $OUTPUT
 RESULT=$?
 # we need to do the chown as we currently cannot select the right user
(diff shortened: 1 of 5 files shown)


rgerhards:
see https://github.com/rsyslog/rsyslog-website/pull/36 for the use case

tools/buildenv/tools/build-doc
```diff
@@ -14,7 +14,10 @@ else
 	export OUTPUT=build
 fi
 cd /rsyslog-doc
-sphinx-build $STRICT -b $FORMAT $SPHINX_EXTRA_OPTS source $OUTPUT
+if [ -e SPHINX_EXTRA_OPTS ]; then
+	export SPHINX_EXTRA_OPTS="$SPHINX_EXTRA_OPTS `cat $SPHINX_EXTRA_OPTS`"
```
deoren: Did you mean this instead?

```diff
-export SPHINX_EXTRA_OPTS="$SPHINX_EXTRA_OPTS `cat $SPHINX_EXTRA_OPTS`"
+export SPHINX_EXTRA_OPTS="$SPHINX_EXTRA_OPTS `cat SPHINX_EXTRA_OPTS`"
```

rgerhards: ohhh... I think so. Just seen that the web site also does not render correctly - might be related ;-)

deoren: @rgerhards Could be. :)

rgerhards: oh, I actually had this on my local copy ... guess I get to the point where juggling with too many things became counter-productive ;-)

PR #525, opened 2018-01-26
mention docker containers as an option for installing rsyslog

rgerhards:

PR #549, opened 2018-02-09
Update subst refs, language tweaks, misc fixes

deoren:
What started as a focused effort turned into me wandering around some related docs and tweaking various items, creating new substitution references, fixing some old HTML doc page references and performing some general cleanup.

- Loosely related to rsyslog/rsyslog-doc#525
- refs rsyslog/rsyslog-doc#452 (a few hard-coded references were fixed here)

@rgerhards Any concerns with this PR?

### DIFF
--- a/source/includes/container_dev_env.inc.rst
+++ b/source/includes/container_dev_env.inc.rst
@@ -1,8 +1,13 @@
 .. note::
 
-   The rsyslog project maintains
-   `rsyslog docker development environment images <https://hub.docker.com/u/rsyslog/>`_.
-   They are ready to go for the build. We recommend them over you own build
-   environment - but you need to make sure it's the same distro/version
-   if you want to copy the binary. Rsyslog docker development images
-   are named `rsyslog_dev_` followed by the distro name.
+   The rsyslog project maintains multiple rsyslog docker development
+   environment images at the |RsyslogDockerHub|_. These images have been
+   configured specifically for use with rsyslog and are recommended
+   over your own build environment. Rsyslog docker development images are
+   named with the ``rsyslog_dev_`` prefix, followed by the distro name.
+
+.. warning::
+
+   If you plan to copy the binary for use outside of the container you need
+   to make sure to use an image of the same distro/version when building
+   rsyslog.
--- a/source/includes/container_dev_env.inc.rst
+++ b/source/includes/container_dev_env.inc.rst
@@ -1,10 +1,11 @@
 .. note::
 
-   The rsyslog project maintains multiple rsyslog docker development
-   environment images at the |RsyslogDockerHub|_. These images have been
-   configured specifically for use with rsyslog and are recommended
-   over your own build environment. Rsyslog docker development images are
-   named with the ``rsyslog_dev_`` prefix, followed by the distro name.
+   The rsyslog project maintains multiple `rsyslog docker development
+   environment images <https://hub.docker.com/u/rsyslog/>`_. These
+   images have been configured specifically for use with rsyslog and are
+   recommended over your own build environment. Rsyslog docker development
+   images are named with the ``rsyslog_dev_`` prefix, followed by the
+   distro name.
 
 .. warning::
 
--- a/source/includes/substitution_definitions.inc.rst
+++ b/source/includes/substitution_definitions.inc.rst
@@ -12,12 +12,23 @@
 .. |FmtAdvancedName| replace:: ``advanced``
 .. |FmtObsoleteName| replace:: ``obsolete legacy``
 
-.. |DockerHubURL| replace:: https://hub.docker.com/u/rsyslog/
+.. |FmtObsoleteDescription| replace:: ``Obsolete Format Equivalents``
+
+.. |RsyslogDockerHub| replace:: official rsyslog Docker Hub
+.. _RsyslogDockerHub: https://hub.docker.com/u/rsyslog/
+
+.. |GitHubDockerProject| replace:: rsyslog-docker project
+.. _GitHubDockerProject: https://github.com/rsyslog/rsyslog-docker
+
+.. |DockerApplianceAlpineDockerHubRepo| replace:: Alpine rsyslog appliance Docker Hub repo
+.. _DockerApplianceAlpineDockerHubRepo: https://hub.docker.com/r/rsyslog/syslog_appliance_alpine/
+
 .. |DockerApplianceAlpineName| replace:: ``rsyslog/syslog_appliance_alpine``
 .. |DockerApplianceAlpineRUN| replace:: ``docker run -ti rsyslog/syslog_appliance_alpine``
 
-.. |FmtObsoleteDescription| replace:: ``Obsolete Format Equivalents``
 
+.. |RsyslogPackageDownloads| replace:: rsyslog package downloads
+.. _RsyslogPackageDownloads: http://www.rsyslog.com/downloads/download-other/
 
 .. |GitHubDocProject| replace:: rsyslog-doc project
 .. _GitHubDocProject: https://github.com/rsyslog/rsyslog-doc/
--- a/source/installation/install_from_source.rst
+++ b/source/installation/install_from_source.rst
@@ -3,7 +3,7 @@ Installing rsyslog from Source
 
 *Written by* `Rainer Gerhards <http://www.adiscon.com/en/people/rainer-gerhards.php>`_
 
-**In this paper, I describe how to install** 
+**In this paper, I describe how to install**
 `rsyslog <http://www.rsyslog.com/>`_. It is intentionally a brief
 step-by-step guide, targeted to those who want to quickly get it up and
 running. For more elaborate information, please consult the rest of the
@@ -12,10 +12,10 @@ running. For more elaborate information, please consult the rest of the
 How to make your life easier...
 -------------------------------
 
-There are :doc:`packages for rsyslog <packages>` available.
-If you use them, you can spare yourself many of the steps below.
-This is highly recommended if there is a package for your distribution
-available.
+In addition to building from source, you can also install |PRODUCT|
+using packages. If you use them, you can spare yourself many of the steps
+below. This is highly recommended if there is a package for your
+distribution available. See :doc:`packages` for instructions.
 
 Steps To Do
 -----------
@@ -25,8 +25,7 @@ Step 1 - Download Software
 
 For obvious reasons, you need to download rsyslog. Here, I assume that
 you use a distribution tarball. If you would like to use a version
-directly from the repository, see `build rsyslog from
-repository <build_from_repo.html>`_ instead.
+directly from the repository, see :doc:`build_from_repo` instead.
 
 Load the most recent build from
 `http://www.rsyslog.com/downloads <http://www.rsyslog.com/downloads>`_.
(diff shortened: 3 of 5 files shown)


source/includes/container_dev_env.inc.rst
```diff
@@ -1,8 +1,13 @@
 .. note::
 
-   The rsyslog project maintains
-   `rsyslog docker development environment images <https://hub.docker.com/u/rsyslog/>`_.
```
rgerhards: I really think we should have the hyperlink in, and so with proper text. This also has todo with getting google juice over the non-official containers... Is it a problem to do the replacement inside a link? Let's discuss concerns.

source/includes/substitution_definitions.inc.rst
```diff
@@ -12,12 +12,23 @@
...
+.. |FmtObsoleteDescription| replace:: ``Obsolete Format Equivalents``
+
+.. |RsyslogDockerHub| replace:: official rsyslog Docker Hub
+.. _RsyslogDockerHub: https://hub.docker.com/u/rsyslog/
+
+.. |GitHubDockerProject| replace:: rsyslog-docker project
+.. _GitHubDockerProject: https://github.com/rsyslog/rsyslog-docker
+
+.. |DockerApplianceAlpineDockerHubRepo| replace:: Alpine rsyslog appliance Docker Hub repo
+.. _DockerApplianceAlpineDockerHubRepo: https://hub.docker.com/r/rsyslog/syslog_appliance_alpine/
```
rgerhards: the names are getting long ... but probably that's what simply is needed...

source/installation/install_from_source.rst
```diff
@@ -12,10 +12,10 @@ running. For more elaborate information, please consult the rest of the
 How to make your life easier...
 -------------------------------
 
-There are :doc:`packages for rsyslog <packages>` available.
```
rgerhards: same comment on hyperlinks with proper text. IMHO we should not sacrifice well-written html for the sake of replacement strings, if that's the issue. But again, I am open to discussion.

source/installation/install_from_source.rst
```diff
@@ -25,8 +25,7 @@ Step 1 - Download Software
 
 For obvious reasons, you need to download rsyslog. Here, I assume that
 you use a distribution tarball. If you would like to use a version
-directly from the repository, see `build rsyslog from
-repository <build_from_repo.html>`_ instead.
```
rgerhards: same as above

source/installation/install_from_source.rst
```diff
@@ -135,22 +134,27 @@ you are upgrading from stock syslogd, /etc/syslog.conf is probably a
 good starting point. Rsyslogd understands stock syslogd syntax, so you
 can simply copy over /etc/syslog.conf to /etc/rsyslog.conf. Note since
 version 3 rsyslog requires to load plug-in modules to perform useful
-work (more about `compatibility notes v3 <v3compatibility.html>`_). To
```
rgerhards: same ;-)

source/installation/packages.rst
```diff
@@ -21,12 +21,17 @@ can help with simple config questions, for anything else we concentrate on
...
 
 The rsyslog project offers current packages for a number of major distributions.
-More information about these can be found at http://www.rsyslog.com/downloads/download-other/.
+More information about these can be found at the |RsyslogPackageDownloads|_
+page.
 
 If you do not find a suitable package for your distribution, there is no
-reason to panic. You can use the :doc:`rsyslog docker containers <rsyslog_docker>`
-instead or
-:doc:`install rsyslog from the source tarball <install_from_source>`.
```
rgerhards: same comment

source/installation/install_from_source.rst
deoren: Here I think I just moved the text. Instead of leading in with "there are packages available", I moved that below with the next section.

Also, a note on the link generation. I think the [sphinx manual](http://www.sphinx-doc.org/en/stable/markup/inline.html) says it better than I probably can:

> :doc:
> Link to the specified document; the document name can be specified in absolute or relative fashion. For example, if the reference :doc:`parrot` occurs in the document sketches/index, then the link refers to sketches/parrot. If the reference is :doc:`/people` or :doc:`../people`, the link refers to people.
> 
> If no explicit link text is given (like usual: :doc:`Monty Python members </people>`), **the link caption will be the title of the given document**.

The part I'm trying to take advantage of is the **the link caption will be the title of the given document** behavior. If the document I am linking too has a good title, I can give a bare reference to the document instead of explicit text that may mask a broken link.

Diff:

```diff
-:doc:`colorful birds that talk back<parrot>`
+:doc:`parrot`
```

In this case the page title is "Installing rsyslog from Package", so I thought a good way to link to the doc was to make use of the page title in the text that was going to link to it. We *could* override the page title and reference it earlier as was done before my edits, but I was attempting to make linking to the "Installing rsyslog from Package" more natural.

The substitution references and definitions is mostly an effort on my part to keep from repeating common phrases, especially if there is a desire to use them more than once. I've already seen in many places in the current version of the docs where content has been repeated and will need to be changed in multiple places. Investing in substitution definitions/references are part of an effort on my part to follow the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

It's a difficult balance, but that is my aim.

source/installation/install_from_source.rst
deoren: This was a broken link. In current docs we shouldn't be referencing reST docs via a .html extension.

source/installation/install_from_source.rst
deoren: This is another broken link.

source/installation/install_from_source.rst
```diff
@@ -135,22 +134,27 @@ you are upgrading from stock syslogd, /etc/syslog.conf is probably a
...
+  $ModLoad immark # provides --MARK-- message capability
+  $ModLoad imudp # provides UDP syslog reception
+  $ModLoad imtcp # provides TCP syslog reception
+  $ModLoad imuxsock # provides support for local system logging
+  $ModLoad imklog # provides kernel logging support
 
 Change rsyslog.conf for any further enhancements you would like to see.
 For example, you can add database writing as outlined in the paper
-"`Writing syslog Data to MySQL <rsyslog_mysql.html>`_\ " (remember you
-need to enable MySQL support during step 2 if you want to do that!).
```
deoren: You didn't say anything about this one, but this is another broken link.

source/installation/install_from_source.rst
```diff
@@ -232,36 +236,3 @@ don't go smooth. In some rare cases, enabling debug logging (-d option)
 in rsyslogd can be helpful. If all fails, go to
 `www.rsyslog.com <http://www.rsyslog.com>`_ and check the forum or
 mailing list for help with your issue.
-
-Housekeeping stuff
```
deoren: This section (lines 235-267) felt like it was in need of pruning. The contact information is already provided in many places, the footer now includes links to multiple "go here for help" resources and the revision history doesn't seem particularly relevant any longer (the Git history is available to anyone who is really interested in the history of a document).

source/installation/packages.rst
```diff
@@ -21,12 +21,17 @@ can help with simple config questions, for anything else we concentrate on
 current versions.
 
 The rsyslog project offers current packages for a number of major distributions.
-More information about these can be found at http://www.rsyslog.com/downloads/download-other/.
+More information about these can be found at the |RsyslogPackageDownloads|_
```
deoren: This actually creates a hyperlink with the text of "rsyslog package downloads" that points to the original URL. I think this is actually a good example of the approach that I think you were recommending that we use.

source/installation/packages.rst
deoren: Here I moved the links to the "see also" section. This was as much an effort to pull out the specific documents and highlight them (via a "see also" section) as much as it was to allow the actual page titles to show through. I personally like the approach of using "see also" sections to bring attention to specific documents as I feel it allows for easy/quick browsing of a doc in order to find related items.

I don't know. I have *my* preferences, but ultimately this is up to you and I understand that.

source/includes/substitution_definitions.inc.rst
deoren: Yeah. Longer names are a trade-off. Sure, they are longer, but they also make more visual sense when you see them while editing a document. With a longer reference, it almost reads like a natural part of the document and less like a programming/markup language construct. Pros and cons.

Another advantage of longer names is the ability to help differentiate between similar substitution references.

source/includes/container_dev_env.inc.rst
deoren: >  Is it a problem to do the replacement inside a link?

I think there is a technical limitation. I recall reading this again recently, but I don't remember where off the top of my head. Let me look for it and I'll respond back here with a reference (providing I can find it).

davidelang:
nowdays, it's more like "you copy anold syslog config and include it into your 
rsyslog.conf", I really doubt there is much of anyone doing things with plain 
syslog any more. This comment made sense in the 2005 timeframe when rsyslog was 
on v3, but should be re-worked and instead say something along the lines of 
"traditional systlog syntax is generally supported as legacy config options"

David Lang

deoren:
> @davidelang: nowdays, it's more like "you copy anold syslog config and include it into your
rsyslog.conf", I really doubt there is much of anyone doing things with plain
syslog any more. This comment made sense in the 2005 timeframe when rsyslog was
on v3, but should be re-worked and instead say something along the lines of
"traditional systlog syntax is generally supported as legacy config options"

Probably so. My goal with editing that portion of the doc (there is still much to do there) was just to fix the broken link.

source/includes/container_dev_env.inc.rst
deoren: While I think we can pull off a replacement inside a link, it does not appear to be as straightforward as it would need to be. The easiest fix here is just to revert to the original link. 

My goal was to put the focus on the rsyslog Docker Hub itself and not the development images. If you feel the focus should be on the images instead, we could alter the subst reference and then rework the other references. This would put the focus on the images instead of the Docker Hub (which I saw as a landing page to learn more about available images).

source/installation/install_from_source.rst
rgerhards: ACK - I need to add some info on why some of these constructs are there. It's actually contributed pages. These is mine, so it is not a problem. Will document the so far non-written policy soon. Good this came up!

source/includes/substitution_definitions.inc.rst
rgerhards: ack - but typing too long names is a pain ;-)

source/includes/container_dev_env.inc.rst
rgerhards: Valid points. I guess that text was already a bit outdated by now. Let's nail this PR and go for the next one.

source/installation/install_from_source.rst
rgerhards: ah, I didn't know it automatically pulls in the page title. So this probably addresses most of my concerns (assuming the title is OK).

source/installation/install_from_source.rst
rgerhards: you are right! I overlooked the html

source/installation/install_from_source.rst
rgerhards: ... towards the end did obviously was not too concentrated on the changes ;-)

source/installation/packages.rst
rgerhards: I am not totally sure, but it sounds reasonable. This reminds me we have a task to make the generation result available, which, as you said, is helpful for such discussions. Let's go ahead with the current proposal.

rgerhards:
I merge, let's  do further improvements in the future. Things have definitely gotten better!

PR #553, opened 2018-02-11
fix docker container: contained rsyslog.com site theme

rgerhards:
The docker container is for the rsyslog-doc project, and should be
agnostic to its users. So it was a bad idea in the first place to
put a website-only dependency into the container. Now removing that.

see also https://github.com/rsyslog/rsyslog-doc/pull/543#issuecomment-364761077

### DIFF
--- a/tools/buildenv/Dockerfile
+++ b/tools/buildenv/Dockerfile
@@ -1,7 +1,7 @@
 FROM	alpine:3.7
 LABEL	maintainer rgerhards@adiscon.com
 RUN	apk add --no-cache py-pip git
-RUN	pip install sphinx sphinx-better-theme
+RUN	pip install sphinx
 RUN	adduser -s /bin/ash -D rsyslog rsyslog \
 	&& echo "rsyslog ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
 WORKDIR	/home/appliance


deoren:
What about running `pip install -r requirements` to match what the main README advises users to do? This way any project dependencies are installed as they expand slowly over time vs having to manually sync the Dockerfile against that file.

rgerhards:
> What about running pip install -r requirements.txt to match what the main README advises users to do? This way any project dependencies are installed as they expand slowly over time vs having to manually sync the Dockerfile against that file.

I don't like that automatism - IMHO a docker file should be hand-crafted to work optimal. Among others, that turns of our ability to cache and so causes long delays in development.

rgerhards:
Let me rephrase a bit: if you would like to actually give it a try, I would not object so. But the resulting docker container should be equally good and the cache usable for things like the core install - everything else causes immense waste of time when working on those things that really change frequently (like scripts).

I would bet that when working on that, you'll probably notice a couple of problems, e.g. files present in the container you do not want to have there, bringing up complexity to remove container layers etc... But I may be wrong.

rgerhards:
I'll merge this as-is so that I can continue with the deployment. Please add a PR for the ```requirements.txt``` task if you think advantages make up for the extra work.

PR #554, opened 2018-02-12
document contribution policy

rgerhards:
this policy is in effect for several years now, but was
never formally written down. So far, it has been communicated
oral, via private mails, the mailing list ... and probably some
other hard to track channels. Now let's officiall "publish" it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,61 @@
+# How to Contribute
+
+Contributing is easy: fork this repo, create a new branch, change the doc
+and then create a pull request.
+
+The system will automatically initiate some tests on your contribution ("CI").
+Please ensure that these checks succeed.
+
+Prior to merge, a review will be made by a maintainer. If questions appear,
+please be ready to address them.
+
+For a successful contribution, it is important to follow up on the pull
+request.
+
+Authors who provided substantial technical content to a documentation page
+can be mentioned (usually be mentioning themselves) on that page. They
+may include a link to their homepage as well as their email address. The
+same applies to rsyslog module authors. By default, the author should be named,
+except if he or she does not want this. Module authors belong into the header
+of the module documentation.
+
+For links, the same policies as laid down below applies.
+
+
+# Contributing Content that already appeared Elsewhere
+
+We are glad you consider adding content that you already published to the
+rsyslog documentation. This makes your content more accessible and helps
+rsyslog users. You are permitted to link back to the original appearance
+of the content.
+
+There are some easy rules for doing so:
+
+- you must be the copyright holder
+- the content must be technical and a useful addition to the existing
+  doc (aka purely promotional or duplicated content is not permitted)
+- there must be no monetization inside the submitted doc content itself
+  (some monetization inside the original content is permitted as
+  elaborated down below)
+- you may provide up to two external links:
+  - one to the author (person or organization)
+  - one to the original place of appearance (blog, company site, ...)
+- additional links may be included if the rsyslog-doc team considers
+  them useful (e.g. another open source library). Links to RFCs, at the
+  official IETF archive are always permitted and unproblematic.
+- links must not lead to malicious, offensive or otherwise inappropriate
+  sites (the validation is done by the rsyslog-doc maintainers and no
+  complaints are accepted)
+
+Note that similar rules exist for content that is destined to appear on
+rsyslog.com. However, we nowadays try to limit documentation type of content
+to the actual documentation.
+
+The same rules also apply to content published via e.g. YouTube.com.
+Monetized content on the original site (and in videos) is permitted as
+long as the main focus is on the technical content. Again, the decision
+of what is appropriate and not is solely made by the rsyslog-doc team.
+
+The core guiding principle is that we would like to have useful content
+which is of help of the user base. We want to make this process as easy
+as possible.


CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
+# How to Contribute
+
+Contributing is easy: just change the doc and create a pull request.
+
+The system will automatically initiate some tests on you contribution ("CI").
+Please ensure that these checks succeed.
```
deoren:
```diff
-you
+your
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
+# How to Contribute
+
+Contributing is easy: just change the doc and create a pull request.
```
deoren: We may need to link to a basic guide describing the need to create a new branch first. I messed up when I first contributed here. At that time I made my changes directly on my master branch, submitted a PR, merged back changes from "upstream" master (which was via a merge commit) and then proceeded to make more changes and then finally tried to create a PR from the resulting mess.

I ended up just nuking my fork and recreating it. After doing some reading I realized where I went wrong and I think I've been OK since. Even if we don't have a solid guide now to link to, maybe make a reference to the required process?

```diff
-Contributing is easy: just change the doc and create a pull request.
+Contributing is easy: fork this repo, create a new branch, change the doc and then create a pull request.
```

It can probably be said simpler and I imagine it will be once we have a small doc to point users to. At that point we can say something like:

```diff
-Contributing is easy: just change the doc and create a pull request.
+Contributing is easy: follow the brief steps on the XYZ page to make changes and submit them for review
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+
+Prior to merge, a review will be made by a maintainer. If questions appear,
+please be ready to address them.
+
+For a successful contribution, it is important to follow up on the pull
+request.
+
+Authors who provided substantial technical content to a documentation page
+can be mentioned (usually be mentioning themselves) on that page. They
+may include a link to their homepage as well as their email address. The
```
deoren: I understand this is a project policy decision, but visually we may need a section intended just for this. Maybe a "credits" section that bundles names, contact info, website, etc. At some point we could even come up with a standard format for the info.

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+rsyslog documentation. This makes your content better accessible and helps
+rsyslog users. You are permitted to link back to the original appearance
+of the content.
+
+There are some easy rules for doing so:
+
+- you must be the copyright holder
+- the content must be technical and a useful addition to the existing
+  doc (aka purely promotional or duplicated content is not permitted)
+- there must be no monetarization inside the submitted doc content itself
```
deoren:
```diff
-monetarization
+monetization
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+  (some monetarization inside the original content is permitted as
```
deoren:
```diff
-monetarization
+monetization
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+  - one to the author (person or organization)
+  - one to the original place of appearance (blog, company site, ...)
+- additional links may be included if the rsyslog-doc team considers
+  them useful (e.g. another open source library). Links to RFCs, at the
+  official IETF archive are always permitted and unproblematic.
+- links must not lead to malicious, offensive or otherwise inappropriate
+  sites (the validation is done by the rsyslog-doc maintainers and no
+  complaints are accepted)
+
+Note that similar rules exists for content that is destined to appear on
```
deoren:
```diff
-exists 
+exist
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+rsyslog.com. However, we nowadays try to limit documentation type of content
+to the actual documentation.
+
+The same rules also apply to content published via e.g. YouTube.com.
+Monetarized content on the original site (and in videos) is permitted as
```
deoren:
```diff
-Monetarized
+Monetization
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+except if he or she does not want this. Module authors belong into the header
+of the module documentation.
+
+For links, the same policies as laid down below applies.
+
+
+# Contributing Content that already appeared Elsewhere
+
+We are glad you consider adding content that you already published to the
+rsyslog documentation. This makes your content better accessible and helps
```
deoren:
```diff
-rsyslog documentation. This makes your content better accessible and helps
+rsyslog documentation. This makes your content more accessible and helps
```

deoren:
> @rgerhards: @deoren would be great if you could have a look if anything catches your eye...

I looked over it and offered some suggestions. One item I wasn't entirely sure about was the use of "monetize" vs "monetarize". I offered a "correction" anyway based on what I believe is right, but certainly double-check me on that.

Overall it's really good to see this content collected like this. Some of the points you cover I was not aware of and am glad I've had a chance to read it before reviewing contributions from others not already part of the team.

CONTRIBUTING.md
rgerhards: This doc here comes up when you do a PR on github. I didn't think about this before, but maybe we should in the future limit it to what you need to know at that instant, with the major of the contribution doc being a separate page. Nevertheless, that's for another shot, let's get the info first in at all.

CONTRIBUTING.md
rgerhards: makes sense - let's work on this later on.

rgerhards:
@deoren thx, I fixed the issues. Will merge now. For the follow up items, let's use new trackers.

CONTRIBUTING.md
deoren: Sure. Just wanted to mention it while I was thinking about it.

refs #451 (bread crumb for later review)

Issue #547, opened 2018-02-08
omstdout: Document new format support


deoren:
See rsyslog/rsyslog#2455

deoren:
Options that I see from looking at the diff:

- `templateName`
- `ensureLFEnding`
- `UseArrayInterface`

I could be missing some, but those stood out after a quick review.

Issue #456, opened 2017-12-22
Create cross-reference of obsolete to advanced format replacements


deoren:
This came up in a recent conversation with @rgerhards. As we go about pruning the older obsolete format content from the documentation it is important that we have a cross-reference in place to guide readers coming from older configurations to the equivalent parameters in the advanced format.

In some cases this is currently handled at the bottom of module docs, but since we're pruning that content as we go we'll need to consider a common index.

We may need to create a mock-up to get the process started just so we're on the same page.

deoren:
related issue: #449

deoren:
Note to self: grep rsyslog source for `eCmdHdlrGoneAway` as one way of helping to identify obsolete legacy directives that are no longer supported.

rgerhards:
> Note to self: grep rsyslog source for eCmdHdlrGoneAway as one way of helping to identify obsolete legacy directives that are no longer supported.

yup - that definitely works (assuming I have all done cleanly... ;)).

deoren:
Some notes from https://github.com/rsyslog/rsyslog-doc/issues/453#issuecomment-353718716:

> @rgerhards: I still think we should remove all obsolete format references and just put them into a single old-to-new mapping page. So it'll be removed from where it originally existed. On the new (single) page, we could add them at the bottom as "removed".
> 
> For new style/script parameters that disappear, I agree to keep them on the same page.

deoren:
> @rgerhards: yup - that definitely works (assuming I have all done cleanly... ;)).

Awesome, I was hoping that would be a good indicator to work from.

deoren:
Update: The work on PR #476 may resolve this in a slightly different way: mention the [`obsolete legacy`](http://www.rsyslog.com/doc/v8-stable/configuration/conf_formats.html) format directives in a small table right with the coverage for the [`advanced`](http://www.rsyslog.com/doc/v8-stable/configuration/conf_formats.html) format parameters.

This keeps the deprecated directive names right with the current parameters for easy reference.

PR #556, opened 2018-02-12
omstdout: rewrite documentation

PascalWithopf:
Module documentation was rewritten in the style.
Additionally parameter with the new config style were added.

closes https://github.com/rsyslog/rsyslog-doc/issues/547

### DIFF
--- a/source/configuration/modules/omstdout.rst
+++ b/source/configuration/modules/omstdout.rst
@@ -1,13 +1,16 @@
+***********************************************
 omstdout: stdout output module (testbench tool)
-===============================================
+***********************************************
 
-**Module Name:    omstdout**
+===========================  ===========================================================================
+**Module Name:**             **omstdout**
+**Author:**                  `Rainer Gerhards <http://rainer.gerhards.net/>`_ <rgerhards@adiscon.com>
+**Available Since:**         4.1.6
+===========================  ===========================================================================
 
-**Author:**\ Rainer Gerhards <rgerhards@adiscon.com>
 
-**Available Since**: 4.1.6
-
-**Description**:
+Purpose
+=======
 
 This module writes any messages that are passed to it to stdout. It
 was developed for the rsyslog test suite. However, there may (limited)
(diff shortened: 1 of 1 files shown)


rgerhards:
@deoren The question @PascalWithopf raises in regard to ```template``` is a very good one. It's actually a problem with almost all of the actions. In obsolete legacy format, the template is (usually ;-)) given by appending ```;TemplateName``` to the action parameters - so, yes, you can set it, but, no, it's not a real parameter.

I admit I do not know yet how to handle this smartly inside the table...

deoren:
It's been a hectic day, but I will try to look at this later today when I have some time.

deoren:
> @rgerhards: @deoren The question @PascalWithopf raises in regard to template is a very good one.

I agree, that is a good question and not necessarily one that has an easy answer.

> @rgerhards: It's actually a problem with almost all of the actions. In obsolete legacy format, the template is (usually ;-)) given by appending ;TemplateName to the action parameters - so, yes, you can set it, but, no, it's not a real parameter.
> 
> I admit I do not know yet how to handle this smartly inside the table...

Maybe it's worth spinning off a separate issue just for this. It's something that many (all?) of the other output module docs will need to cover eventually, so maybe it's good to handle that decision in one place.

source/configuration/modules/omstdout.rst
```diff
@@ -16,21 +19,96 @@ the quality of this module as we do not expect it to be used in real
...
+   module(load="omstdout")
+   action(type="omstdout")
+
+
+Example 2
+---------
+
+The following sample will write syslog messages to stdout, using a template
+and input from imtcp.
+
```
deoren: I'm probably overlooking it, but how does imtcp factor into the example?

PascalWithopf: Sry, I took imtcp out of the example to keep it as short as possible but forgot to change the description

deoren:
refs: #456, #451

source/configuration/modules/omstdout.rst
```diff
@@ -16,21 +19,95 @@ the quality of this module as we do not expect it to be used in real
...
-Note: configuration directive names are case-insensitive.
+Configuration
+=============
+
+.. note::
+
+   Parameter names are case-insensitive.
+
+
+Module Parameter
```
deoren:
```diff
-Module Parameter
+Module Parameters
```

Should this be plural?

source/configuration/modules/omstdout.rst
```diff
@@ -16,21 +19,95 @@ the quality of this module as we do not expect it to be used in real
...
+----------------
+
+none
+
+
+Action Parameter
```
deoren: Same here.

PR #563, opened 2018-02-13
im3195: rewrite documentation

PascalWithopf:

### DIFF
--- a/source/configuration/modules/im3195.rst
+++ b/source/configuration/modules/im3195.rst
@@ -1,25 +1,54 @@
+****************************
 im3195: RFC3195 Input Module
-============================
+****************************
+
+===========================  ===========================================================================
+**Module Name:**             **im3195**
+**Author:**                  `Rainer Gerhards <http://rainer.gerhards.net/>`_ <rgerhards@adiscon.com>
+===========================  ===========================================================================
+
+
+Purpose
+=======
 
 Receives syslog messages via RFC 3195. The RAW profile is fully
 implemented and the COOKED profile is provided in an experimental state.
 This module uses `liblogging <http://www.liblogging.org>`_ for the
 actual protocol handling.
 
-**Author:**\ Rainer Gerhards <rgerhards@adiscon.com>
 
-Parameters
-------------------------
+Configuration Parameters
+========================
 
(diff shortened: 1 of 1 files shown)


deoren:
@PascalWithopf Do you think it is worth the effort to convert the [`obsolete legacy`](http://www.rsyslog.com/doc/v8-stable/configuration/conf_formats.html) configuration format example over to the current format? From what I can tell this module isn't built by default and after reviewing the notes it doesn't appear that it will be going forward.

deoren:
I found some issues with the recent changes to imrelp.rst, impstats.rst and imptcp.rst:

```
/home/deoren/dev/rsyslog/rsyslog-doc/source/configuration/modules/impstats.rst:291: WARNING: duplicate label statistic-counter, other instance in /home/deoren/dev/rsyslog/rsyslog-doc/source/configuration/modules/imrelp.rst
/home/deoren/dev/rsyslog/rsyslog-doc/source/configuration/modules/imptcp.rst:540: WARNING: duplicate label statistic-counter, other instance in /home/deoren/dev/rsyslog/rsyslog-doc/source/configuration/modules/impstats.rst
/home/deoren/dev/rsyslog/rsyslog-doc/source/configuration/modules/imrelp.rst:343: WARNING: duplicate label statistic-counter, other instance in /home/deoren/dev/rsyslog/rsyslog-doc/source/configuration/modules/imptcp.rst
```

The conflict was not immediately obvious from CI since each PR stood alone and was not rebased against master between each merge (as a result the conflict was not seen).

This will require updating all module docs using the new format to use a unique label for the stats counter section.

I'll submit a PR for those module docs and imuxsock.rst soon.

PascalWithopf:
@deoren I would leave the example as it is, since the one parameter it has is only available in the legacy format.

deoren:
> @PascalWithopf: @deoren I would leave the example as it is, since the one parameter it has is only available in the legacy format.

Good point, I missed that remark when I glanced over the doc.

deoren:
> @deoren: This will require updating all module docs using the new format to use a unique label for the stats counter section.

Turns out that only the module docs already merged were affected.

PR #570, opened 2018-02-13
imkmsg: rewrite documentation

PascalWithopf:

### DIFF
--- a/source/configuration/modules/imkmsg.rst
+++ b/source/configuration/modules/imkmsg.rst
@@ -1,12 +1,16 @@
+**********************************
 imkmsg: /dev/kmsg Log Input Module
-==================================
+**********************************
 
-**Module Name:    imkmsg**
+===========================  ===========================================================================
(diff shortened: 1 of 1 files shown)


deoren:
Do you know if this module is deprecated or is it just not built by default? When I attempted to load it via `$ModLoad imkmsg` rsyslogd was unable to find the module.

rgerhards:
I just checked, it is in the contrib tree. So I guess @PascalWithopf has just changed the doc format.

deoren:
@rgerhards @PascalWithopf Is it worth noting that the module is not built by default?

rgerhards:
Very, very slippery slope: each distro decides what they build and what the
packaging structure is - quite different btw...


Sent from phone, thus brief.

deoren:
> @rgerhards: Very, very slippery slope: each distro decides what they build and what the
packaging structure is - quite different btw...

Yeah, I can see that; there may be some distros that actually do build it.

PR #573, opened 2018-02-15
impstats: rewrite documentation

PascalWithopf:

### DIFF
--- a/source/configuration/modules/impstats.rst
+++ b/source/configuration/modules/impstats.rst
@@ -1,7 +1,15 @@
+***********************************************************
 impstats: Generate Periodic Statistics of Internal Counters
 ***********************************************************
 
-**Author:**\ Rainer Gerhards <rgerhards@adiscon.com>
+===========================  ===========================================================================
+**Module Name:**             **impstats**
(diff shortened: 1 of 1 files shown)


deoren:
@PascalWithopf Re the Caveats / Known Bugs section: 

> This module MUST be loaded right at the top of rsyslog.conf, otherwise stats may not get turned on in all places.

Is that still relevant for the [`advanced`](http://www.rsyslog.com/doc/v8-stable/configuration/conf_formats.html) configuration format?

rgerhards:
I am 95% sure it's right, but for 100% a code review is required.

deoren:
> @rgerhards: I am 95% sure it's right, but for 100% a code review is required.

That the placement at the top of the conf is required or that it is not required?

rgerhards:
I think it needs to go at the top -- IIRC impstats switches a know which is read by other modules at load time. So I would remove that statement only after a code review.

deoren:
> @rgerhards: I think it needs to go at the top -- IIRC impstats switches a know which is read by other modules at load time. So I would remove that statement only after a code review.

ACK.

PR #578, opened 2018-02-16
Add back project prefix to dev build string for epub builds

deoren:
This was unintentionally left off in a recent edit to simplify the default dev build string.


deoren:
The CI build may fail due to warnings generated from recent changes to the module docs (duplicate label warnings).

deoren:
> @deoren: The CI build may fail due to warnings generated from recent changes to the module docs (duplicate label warnings).

Both builds failed. I may merge anyway since the failure is not directly related to this change, but instead to the changes merged the previous two days.

deoren:
Rebased off of master to get the fix for the duplicate labels.

Issue #583, opened 2018-02-21
imfile: 'Facility' and 'Severity' parameters have questionable defaults


deoren:
I think this is probably just a mistake that crept in with recent work, but as of d90f47b59a7128ffe8bf467ec66a8b925db2b449 these items need review:

- Facility
    - Table says '128' is default value
    - Text body for that section lists '16' for 'local0' default. I would think that 'local0' would be the value listed instead of the integer value (which does not appear to be correct based on the text block)

- Severity
    - Table says '5' is default value
    - Text body for that section lists 'notice' as default (which *does* correspond to 5 according to [this page](https://en.wikipedia.org/wiki/Syslog#Severity_level)), but table lists an integer. For consistency this probably should be updated to a text value

- "Type" column in the table for both Facility and Severity does not include the actual parameter type (integer, string, etc). Instead, it just lists the parameter name again as the type (which arguably *is* true, but is still not the data type).
    - The description for each parameter makes clear that both numerical and text form are supported, but notes that the textual form is recommended. We should update the type to list both integer and string and should update the 'default' column to list the textual default value

PR #585, opened 2018-02-21
imfile: Update facility and severity parameters

deoren:
- Note that the type is integer or string, with string values preferred

- Emphasized that both severity and facility are applied to messages
  read from the specified file (previously emphasized "lines read")

- Add "see also" for additional Facility and Severity info
  by way of Wikipedia Syslog page

- Update default column in both tables to use textual version
    - corrected default for Facility parameter

closes rsyslog/rsyslog-doc#583

### DIFF
--- a/source/configuration/modules/imfile.rst
+++ b/source/configuration/modules/imfile.rst
@@ -225,11 +225,15 @@ Facility
    :widths: auto
    :class: parameter-table
 
-   "facility", "128", "no", "``$InputFileFacility``"
+   "integer or string (preferred)", "local0", "no", "``$InputFileFacility``"
 
-The syslog facility to be assigned to lines read. Can be specified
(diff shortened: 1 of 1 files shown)


deoren:
@davidelang Any concerns with these changes?

davidelang:
looks good

FYI, I think the reason why it said 128 before is that the facility gets 
multiplied by 8 and added to the severity to produce the priority value, so 
facility 16 becomes 128 in the pri calculation


David Lang

deoren:
> @davidelang: looks good

Thanks for looking over it.

> FYI, I think the reason why it said 128 before is that the facility gets
> multiplied by 8 and added to the severity to produce the priority value, so
> facility 16 becomes 128 in the pri calculation

Ah, OK. I can see how the range could be mistaken because of that.

PR #586, opened 2018-02-23
imudp: rewrite documentation

PascalWithopf:


deoren:
Note to self: same discussion points for this as with #587.

PascalWithopf:
@deoren I have added the Additional Resources section here too.

deoren:
@PascalWithopf Thanks.

Same question on this one as with https://github.com/rsyslog/rsyslog-doc/pull/593#issuecomment-369331387: `word` vs `string` for parameter types.

PR #587, opened 2018-02-23
imtcp: rewrite documentation

PascalWithopf:

### DIFF
--- a/source/configuration/modules/imtcp.rst
+++ b/source/configuration/modules/imtcp.rst
@@ -1,227 +1,424 @@
+******************************
 imtcp: TCP Syslog Input Module
-==============================
+******************************
+
+===========================  ===========================================================================
+**Module Name:**             **imtcp**
+**Author:**                  `Rainer Gerhards <http://rainer.gerhards.net/>`_ <rgerhards@adiscon.com>
+===========================  ===========================================================================
+
+
+Purpose
+=======
 
 Provides the ability to receive syslog messages via TCP. Encryption is
 natively provided by selecting the approprioate network stream driver
 and can also be provided by using `stunnel <rsyslog_stunnel.html>`_ (an
 alternative is the use the `imgssapi <imgssapi.html>`_ module).
 
-**Author:**\ Rainer Gerhards <rgerhards@adiscon.com>
+
+Notable Features
+================
+
+- :ref:`imtcp-statistic-counter`
+
 
 Configuration Parameters
-------------------------
+========================
+
+.. note::
+
+   Parameter names are case-insensitive.
+
 
 Module Parameters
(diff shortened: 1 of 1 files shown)


deoren:
@PascalWithopf The only thing that sticks out at the moment is the link to the video tutorial in the "Notable Features" section. I think it's good to reference external resources, I'm just not sure if that is the best spot for it.

I'm guilty of removing the "See Also" section from the v8 imuxsock module doc (which became a template for recent work), but I wonder whether restoring it (or a section of similar purpose) would be the way to go.

For example, here is a v7 copy of the imuxsock doc:

http://www.rsyslog.com/doc/v7-stable/configuration/modules/imuxsock.html

I ended up finding ways to collapse the "See Also" section in the v8 doc because at the time it felt like having those specific items closer to the main topic (or in the case of Solaris support putting, it into the Caveats section) was a better approach. 

What do you think of a section named "Additional Resources"? Not crazy about the name, but I think keeping "extra" resource links separate from specific points of interest about the module itself would be useful.

@davidelang Any thoughts on this?

davidelang:
I need to see an example of what you are suggesting. I'm not understanding it 
clearly
k

PascalWithopf:
@deoren Yeah, I wasn't so sure about putting it there either.

Adding a section for additional resources seems like a good solution. I will change it.

deoren:
> @PascalWithopf: @deoren Yeah, I wasn't so sure about putting it there either.
> 
> Adding a section for additional resources seems like a good solution. I will change it.

What do you think about having the new section towards the bottom of the page? That way it serves as a, "now that you've finished reading this page, go here for more info" section.

deoren:
> @davidelang: I need to see an example of what you are suggesting. I'm not understanding it clearly

I was referring to creating a dedicated section intended for additional reading to the bottom of the doc. I also mentioned that there used to be such a section on the `imuxsock` module doc until I merged the relevant links into the "Caveats/Known bugs" section and the parameter sections.

If the external reference were specific to one of the parameters, my personal preference (though I can't say how good it is) would be to use a `seealso` directive to highlight the extra material right at the applicable parameter section.

In this case, the extra material is a page on the main site that provides a tutorial video for routing messages into separate files (using the [`obsolete legacy`](http://www.rsyslog.com/doc/v8-stable/configuration/conf_formats.html) configuration format.). Because of that, it could go in a section at at the top to draw the most focus (which to me sort of implies, "watch/read this first") or at the bottom as additional material after reviewing the page.

PascalWithopf:
@deoren Yes, it should be good like this now.

deoren:
> @PascalWithopf @deoren Yes, it should be good like this now.

Thanks!

PR #593, opened 2018-02-28
omelasticsearch: rewrite documentation

PascalWithopf:

### DIFF
--- a/source/configuration/modules/omelasticsearch.rst
+++ b/source/configuration/modules/omelasticsearch.rst
@@ -1,139 +1,284 @@
+********************************************
 omelasticsearch: Elasticsearch Output Module
-============================================
+********************************************
 
-**Module Name:    omelasticsearch**
+===========================  ===========================================================================
+**Module Name:**             **omelasticsearch**
+**Author:**                  `Rainer Gerhards <http://rainer.gerhards.net/>`_ <rgerhards@adiscon.com>
+===========================  ===========================================================================
 
-**Author:**\ Rainer Gerhards <rgerhards@adiscon.com>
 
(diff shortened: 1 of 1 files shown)


deoren:
I looked over the doc and it looks pretty good. I had a question about the "type" column values though. Do you see "word" as more descriptive than "string"? I ask because I think we may have a mix of docs having one or the other.

I don't think I caught it at the time, but the `imrelp` doc is one example where both are used in the same doc.

Example: See the `Ruleset` and `Port` parameters.

PascalWithopf:
@deoren 
String and word are different parameter types. When you look into the module code you can see these types: (from omfile)
```
	{ "file", eCmdHdlrString, 0 },
	{ "dynafile", eCmdHdlrString, 0 },
	{ "sig.provider", eCmdHdlrGetWord, 0 },
	{ "cry.provider", eCmdHdlrGetWord, 0 },
```
The type string is a normal string.
The type word means that only one word is allowed without spaces.

deoren:
> @PascalWithopf: @deoren 
>
> String and word are different parameter types. When you look into the module code you can see these types: (from omfile)
> ```
> 	{ "file", eCmdHdlrString, 0 },
> 	{ "dynafile", eCmdHdlrString, 0 },
> 	{ "sig.provider", eCmdHdlrGetWord, 0 },
> 	{ "cry.provider", eCmdHdlrGetWord, 0 },
> ```
> The type string is a normal string.
> The type word means that only one word is allowed without spaces.

Thank you explaining that; I learned something new. I'll go back and review the work I've done to make sure I've not mistakenly marked a parameter as the wrong type.

PR #596, opened 2018-03-01
omfwd: rewrite documentation

PascalWithopf:


source/configuration/modules/omfwd.rst
```diff
@@ -1,385 +1,540 @@
...
+
+.. csv-table::
+   :header: "type", "default", "mandatory", "|FmtObsoleteName| directive"
+   :widths: auto
+   :class: parameter-table
+
+   "word", "udp", "no", "none"
+
+Type of protocol to use for forwarding. Note that \`\`tcp'' means
+both legacy plain tcp syslog as well as RFC5425-based TCL-encrypted
```
deoren:
```diff
-TCL-encrypted
+TLS-encrypted
```

Is that intended instead?

source/configuration/modules/omfwd.rst
```diff
@@ -1,385 +1,540 @@
...
+syslog. Which one is selected depends on the protocol drivers set
+before the action commend. Note that as of 6.3.6, there is no way to
```
deoren: > Which one is selected **depends on the protocol drivers set before the action commend**. 

Looks like a mispelling and possibly a reference to the [`obsolete legacy`](http://www.rsyslog.com/doc/v8-stable/configuration/conf_formats.html) configuration format having specific directive ordering requirements.

> Note that as of 6.3.6, there is no way to specify this within the action itself.

Does the [`advanced`](http://www.rsyslog.com/doc/v8-stable/configuration/conf_formats.html) configuration format provide that functionality now?

source/configuration/modules/omfwd.rst
PascalWithopf: No I didn't notice this error.

source/configuration/modules/omfwd.rst
PascalWithopf: I have changed the description for that parameter. Yes it does with the StreamDriver parameter.

PR #597, opened 2018-03-01
omhdfs: rewrite documentation

PascalWithopf:

### DIFF
--- a/source/configuration/modules/omhdfs.rst
+++ b/source/configuration/modules/omhdfs.rst
@@ -1,33 +1,72 @@
+***************************************
 omhdfs: Hadoop Filesystem Output Module
-=======================================
+***************************************
 
-**Module Name:    omhdfs**
+===========================  ===========================================================================
+**Module Name:**             **omhdfs**
+**Author:**                  `Rainer Gerhards <http://rainer.gerhards.net/>`_ <rgerhards@adiscon.com>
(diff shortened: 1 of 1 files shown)


deoren:
@PascalWithopf What do you think about converting the old directive entries to separate sections with parameter tables?

Is this one of those cases where it's just not worth the time/trouble to add those details, perhaps because of the low use for the module? 

I don't say that disparagingly of the work done, just that it appears from the content in the doc (this is probably my first time really looking at it) that this module is probably not in heavy use due to the setup requirements.

PascalWithopf:
@deoren I think in this case we shouldn't create another section because only old directives are available and they are the parameters needed when someone wants to use the module.

But we can still put them in a table. I created a table for every parameter but we could also put them into one table. Thought then the description would need space in the table as well.

deoren:
> @PascalWithopf: @deoren I think in this case we shouldn't create another section because only old directives are available and they are the parameters needed when someone wants to use the module.

Fair enough. One upside with a separate section per module is the ability to link right to a parameter, but with the entire parameter list being as small as it is that shouldn't be an issue here.

> @PascalWithopf: But we can still put them in a table. I created a table for every parameter but we could also put them into one table. Thought then the description would need space in the table as well.

I think what you've done has worked out nicely. This looks ready to merge to me.

Thanks for your work on this.

PR #600, opened 2018-03-01
omhttpfs: rewrite documentation

PascalWithopf:

### DIFF
--- a/source/configuration/modules/omhttpfs.rst
+++ b/source/configuration/modules/omhttpfs.rst
@@ -1,5 +1,6 @@
+*************************************
 omhttpfs: Hadoop HTTPFS Output Module
-=====================================
+*************************************
 
 ===========================  ===========================================================================
 **Module Name:**             **omhttpfs**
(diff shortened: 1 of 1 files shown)


deoren:
@PascalWithopf Not an objection, just a musing of sorts:

Is there value in listing the default value for a parameter in the description text for a parameter if it is already included in the table just above it? 

* Pro: help make the default more obvious
* Con: should a default value be changed we will have two places to update it

PascalWithopf:
@deoren I think this should be judged on an individual basis. Sometimes it is unnecessary to do so and then I would take it out of the description, but there are also cases were the default is explained or further information is given, then I would leave it in there.

source/configuration/modules/omhttpfs.rst
```diff
@@ -1,58 +1,149 @@
...
+^^^^
+
+.. csv-table::
+   :header: "type", "default", "mandatory", "|FmtObsoleteName| directive"
+   :widths: auto
+   :class: parameter-table
+
+   "word", "127.0.0.1", "no", "none"
+
+HttpFS server host. Default: *127.0.0.1*
```
deoren: @PascalWithopf Here is an example which prompted my question.

deoren:
> @PascalWithopf: @deoren I think this should be judged on an individual basis. Sometimes it is unnecessary to do so and then I would take it out of the description, but there are also cases were the default is explained or further information is given, then I would leave it in there.

Acknowledged. That makes sense. I picked out one example from the doc rewrite to accompany my question (I should have done that to start with).

PascalWithopf:
@deoren I looked over the parameters and deleted all unnecessary repetitions.

deoren:
@PascalWithopf Thanks!

PR #603, opened 2018-03-02
omkafka: rewrite documentation

PascalWithopf:

### DIFF
--- a/source/configuration/modules/omkafka.rst
+++ b/source/configuration/modules/omkafka.rst
@@ -29,32 +29,7 @@ Action Parameters
 
 Note that omkafka supports some *Array*-type parameters. While the parameter
 name can only be set once, it is possible to set multiple values with that
-single parameter.
-
-For example, to select "snappy" compression, you can use
-
-.. code-block:: none
-
-   action(type="omkafka" topic="mytopic" confParam="compression.codec=snappy")
-
-
-which is equivalent to
-
-.. code-block:: none
-
-   action(type="omkafka" topic="mytopic" confParam=["compression.codec=snappy"])
-
-
-To specify multiple values, just use the bracket notation and create a
-comma-delimited list of values as shown here:
-
-.. code-block:: none
-
-   action(type="omkafka" topic="mytopic"
-          confParam=["compression.codec=snappy",
-	             "socket.timeout.ms=5",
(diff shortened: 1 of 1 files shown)


deoren:
@PascalWithopf Pushed to your branch based on prior permission to do so. Please let me know whether that was intended instead as a one-time thing. 

What do you think of the changes?

PascalWithopf:
@deoren thanks, you can keep pushing your changes, that makes it easier.
I changed the titles of the Example section to match those of the other documentations.

deoren:
> @PascalWithopf: @deoren thanks, you can keep pushing your changes, that makes it easier.

Great, it is working well thus far for me as well.

> @PascalWithopf: I changed the titles of the Example section to match those of the other documentations.

Good idea. I took another look and saw the 'Example 1' section still in place. I pruned it with another push to your branch. Does that work?

If so it looks like this doc is probably ready for merging.

PascalWithopf:
@deoren That is probably a misunderstanding. When I said that I changed the title of the Example section, I added the "Example 1" section.

In this case it looks like it's not necessary, but I when there are multiple examples I think it is useful, especially in a case like this where the example is split into two parts.

deoren:
> @PascalWithopf: @deoren That is probably a misunderstanding. When I said that I changed the title of the Example section, I added the "Example 1" section.
> 
> In this case it looks like it's not necessary, but I when there are multiple examples I think it is useful, especially in a case like this where the example is split into two parts.

Are you saying that the named example sections are not desired, or just for situations where an example is meant to be provided in parts?

PascalWithopf:
@deoren I think the named example sections are a good idea. But in this case I think one title for the whole example is a good idea, because it is related.

deoren:
> @PascalWithopf: @deoren I think the named example sections are a good idea. But in this case I think one title for the whole example is a good idea, because it is related.

Going that route, do you see the fix being to just remove all parent sections so that the single example entry fits below the "Examples" heading?

PascalWithopf:
@deoren I would do it like my commit just now.

deoren:
> @PascalWithopf: @deoren I would do it like seen in my commit just now.

I see what you mean now, group the two items under a common header so it is one logical example. That works.

PR #607, opened 2018-03-05
ommongodb: rewrite documentation

PascalWithopf:

### DIFF
--- a/source/configuration/modules/ommongodb.rst
+++ b/source/configuration/modules/ommongodb.rst
@@ -1,56 +1,121 @@
+********************************
 ommongodb: MongoDB Output Module
-================================
+********************************
 
-**Module Name:    ommongodb**
+===========================  ===========================================================================
+**Module Name:**             **ommongodb**
+**Author:**                  `Rainer Gerhards <http://rainer.gerhards.net/>`_ <rgerhards@adiscon.com>
+===========================  ===========================================================================
 
-**Author:**\ Rainer Gerhards <rgerhards@adiscon.com>
 
-**Description**:
+Purpose
+=======
 
 This module provides native support for logging to MongoDB.
 
-**Action Parameters**:
 
-Note: parameter names are case-insensitive.
+Configuration Parameters
+========================
 
--  **uristr**
-   MongoDB connexion string, as defined by the MongoDB String URI Format (See: https://docs.mongodb.com/manual/reference/connection-string/). If uristr is defined, following directives will be ignored: server, serverport, uid, pwd.
--  **ssl_cert**
-   Absolute path to the X509 certificate you want to use for TLS client authentication. This is optional.
--  **ssl_ca**
-   Absolute path to the trusted X509 CA certificate that signed the mongoDB server certificate. This is optional.
--  **server** (deprecated, use "uristr" instead)
-   Name or address of the MongoDB server
--  **serverport** (deprecated, use "uristr" instead)
-   Permits to select a non-standard port for the MongoDB server. The
-   default is 0, which means the system default port is used. There is
-   no need to specify this parameter unless you know the server is
-   running on a non-standard listen port.
--  **db**
-   Database to use
--  **collection**
-   Collection to use
--  **uid** (deprecated, use "uristr" instead)
-   logon userid used to connect to server. Must have proper permissions.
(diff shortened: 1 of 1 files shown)


source/configuration/modules/ommongodb.rst
```diff
@@ -1,56 +1,176 @@
...
+   :header: "type", "default", "mandatory", "|FmtObsoleteName| directive"
+   :widths: auto
+   :class: parameter-table
+
+   "word", "none", "no", "none"
+
+Absolute path to the trusted X509 CA certificate that signed the mongoDB server certificate. This is optional.
+
+
+(Server) deprecated, use "UriStr" instead
```
PascalWithopf: How should I handle deprecated configuration parameter in the new style?

deoren:
> @PascalWithopf: How should I handle deprecated configuration parameter in the new style?

Good question, and sorry for the radio silence (time is scarce for me right now).

Perhaps handle it like with the `imfile` doc?

http://www.rsyslog.com/doc/v8-stable/configuration/modules/imfile.html#deprecated-parameters

I recall a discussion of having a central list vs keeping a list on each module page, but since we've got at least one other module doc with the deprecated parameters on the same page I vote that we go with that approach for now.

PascalWithopf:
@deoren I will change it like imfile and if we want to create a central list we can still do this later.

deoren:
> @PascalWithopf: @deoren I will change it like imfile and if we want to create a central list we can still do this later.

Acknowledged.

Are all four of these parameters deprecated?

- Server
- ServerPort (shown in 55942da351e5e8bcb945b2ed937c6ba443f9287d as `ServerPorted`)
- UID
- PWD

I ask because the "Deprecated Parameters" section is just above the "Action Parameters" section where those four parameters are listed in 55942da351e5e8bcb945b2ed937c6ba443f9287d.

PascalWithopf:
@deoren Yes they are all deprecated. They are replaced by the parameter `UriStr`.
`ServerPorted` is a typo.
And I created the Deprecated Parameters section at the end of the file separate of the other parameters.

deoren:
> @PascalWithopf: @deoren Yes they are all deprecated. They are replaced by the parameter UriStr.
> ServerPorted is a typo.
> And I created the Deprecated Parameters section at the end of the file separate of the other parameters.

I see now. When I glanced quickly it appeared that the Deprecated Parameters section was sandwiched between two other sections, but I see now that the 'Action Parameters' section there is a sub-section. Do we need that subsection? I wonder if it's perhaps subdividing what is already a short list of content.

deoren:
@PascalWithopf I saw your agreement on a different PR re pushing back to your branch, so I took a stab at introducing some changes here. Please let me know what you think.

PascalWithopf:
@deoren I think you are right, there is no reason to divide the Deprecated Parameters section. Thanks for the help.

deoren:
> @PascalWithopf: @deoren I think you are right, there is no reason to divide the Deprecated Parameters section. Thanks for the help.

You are welcome. Your remarks on a different PR re the section names for the examples reminded me to take a closer look here. I adjusted the names for those sections in an effort to match the others.

Any concerns?

PascalWithopf:
Same as here https://github.com/rsyslog/rsyslog-doc/pull/603#issuecomment-373403028

deoren:
> > @PascalWithopf: @deoren I think you are right, there is no reason to divide the Deprecated Parameters section. Thanks for the help.
> 
> You are welcome. Your remarks on a different PR re the section names for the examples reminded me to take a closer look here. I adjusted the names for those sections in an effort to match the others.
> 
> Any concerns?
>
> @PascalWithopf: Same as here https://github.com/rsyslog/rsyslog-doc/pull/603#issuecomment-373403028

Do you wish to keep the generic names instead of named examples? I prefer the latter because if I'm looking at the Table of Contents in the side bar I can jump right to the example I'm looking for. I can see how it may be useful to refer to a specific generically named example from within the docs (perhaps from a parameter description), but that can be worked around by using a label reference.

PascalWithopf:
@deoren I have changed the example headings as we discussed in https://github.com/rsyslog/rsyslog-doc/pull/603

deoren:
> @PascalWithopf: @deoren I have changed the example headings as we discussed in #603

ACK.

Issue #611, opened 2018-03-07
Is specifying the queue.filename parameter deprecated or discouraged?


deoren:
I thought I recall reading somewhere where it was (that it should be handled automatically now), but after searching I found that it is the stateFile parameter for the `imfile` module which is deprecated.

* Is the `queue.filename` parameter required?
* Are there plans to deprecate the `queue.filename` parameter?

refs:

- http://www.rsyslog.com/doc/master/rainerscript/queue_parameters.html
- http://www.rsyslog.com/doc/master/configuration/modules/imfile.html

davidelang:
It is required, there are no plans to eliminate it

specifying a filename changes behavior (makes the queue disk assisted), so it's 
not just overriding a default failename.

In some ways it's unexpected side effects territory.

David Lang

deoren:
@davidelang Thanks for the info. I'll try to work that in.

deoren:
> @davidelang: It is required, there are no plans to eliminate it
> 
> specifying a filename changes behavior (makes the queue disk assisted), so it's
> not just overriding a default failename.
> 
> In some ways it's unexpected side effects territory.

I thought I understood, but going to make the changes to the doc I read over the second sentence and realized that I perhaps I don't. If the `queue.filename` parameter is not specified, will queues still function, perhaps just in one of the other modes?

davidelang:
On Wed, 7 Mar 2018, Deoren Moor wrote:

>> @davidelang: It is required, there are no plans to eliminate it
>>
>> specifying a filename changes behavior (makes the queue disk assisted), so it's
>> not just overriding a default failename.
>>
>> In some ways it's unexpected side effects territory.
>
> I thought I understood, but going to make the changes to the doc I read over 
> the second sentence and realized that I perhaps I don't.

> If the `queue.filename` parameter is not specified, will queues still 
> function, perhaps just in one of the other modes?

Exactly, the only difference between an in-memory array queue and a memory + 
disk queue is that the filename is specified

so specifying a filename isn't just changing from the default filename, it's 
turning on the disk assisted queue functionality as well.

It's not unexpected to rsyslog, but is unexpected for users.

deoren:
@davidelang Thanks for taking the time to explain that, I appreciate it.

PR #619, opened 2018-03-09
queue parameters: State need for queue.filename

deoren:
Without enabling the option, disk-assisted queue functionality is not available.

- Credit to  @davidelang for clarifying this.
- See also: "Understanding rsyslog Queues"

closes rsyslog/rsyslog-doc#611


rgerhards:
LGTM

Issue #622, opened 2018-03-09
Update community resource links in footer include file: remove forums, push mailing list to the top and include StackExchange link


deoren:
As the subject says.

Snippet from a mailing list discussion:

> So how about this?
> 
> - do not mention forum any longer
> - say "github is for questions"
> - say "SE is experimentally for questions, but less likely to draw
> answers from team"
> - keep "mailing list is great to ask questions"
> 
> update this in doc, site etc.
> 
> If nobody violently objects by monday morning, let's implement the
> changes. We may not get everything immediately, but that's it. I'll
> see that I can change the forum to read-only, but that needs to go to
> web guys and is not top priority.
> 
> Rainer

This ticket is for the doc update portion described here.

PR #624, opened 2018-03-09
Update references to support resources

deoren:
This is per a recent mailing list conversation.

closes rsyslog/rsyslog-doc#622

### DIFF
--- a/source/includes/footer.inc.rst
+++ b/source/includes/footer.inc.rst
@@ -5,9 +5,12 @@
    Help with configuring/using |PRODUCT|:
 
    - |RsyslogOfficialMailingList|_ - best route for general questions
-   - GitHub: |GitHubSourceProject|_ - best route for detailed questions and
-     for reporting issues that are believed to be bugs with |PRODUCT|
-   - |RsyslogStackOverflow|_ - experimental support from rsyslog community
+
+   - GitHub: |GitHubSourceProject|_ - detailed questions, reporting issues
+     that are believed to be bugs with |PRODUCT|
+
+   - Stack Exchange (|RsyslogStackExchangeView|_, |RsyslogStackExchangeAsk|_)
+     - experimental support from rsyslog community
 
 .. seealso::
 
--- a/source/includes/footer.inc.rst
+++ b/source/includes/footer.inc.rst
@@ -2,15 +2,17 @@
 
 .. seealso::
 
-   If you would like to contribute to these docs, but are unsure where to
-   start, please see the |GitHubDocProjectREADME|_ for an overview of the
-   process. If you would like to contribute to the main source project,
-   please review the contribution guidelines listed in the
-   |GitHubSourceProjectREADME|_.
+   Help with configuring/using |PRODUCT|:
 
-   If you have a question about these docs or |PRODUCT| in general, please
-   see the following resources:
+   - |RsyslogOfficialMailingList|_ - best route for general questions
+   - GitHub: |GitHubSourceProject|_ - best route for detailed questions and
+     for reporting issues that are believed to be bugs with |PRODUCT|
+   - |RsyslogStackOverflow|_ - least activity by dev team, but generally
+     good support from the wider tech community
 
-   - |RsyslogOfficialForums|_
-   - |RsyslogOfficialMailingList|_
-   - GitHub: |GitHubDocProject|_, |GitHubSourceProject|_
+.. seealso::
+
+   Contributing to |PRODUCT|:
+
+   - Source project: |GitHubSourceProjectREADME|_.
+   - Documentation: |GitHubDocProjectREADME|_
--- a/source/includes/footer.inc.rst
+++ b/source/includes/footer.inc.rst
@@ -7,8 +7,7 @@
    - |RsyslogOfficialMailingList|_ - best route for general questions
    - GitHub: |GitHubSourceProject|_ - best route for detailed questions and
      for reporting issues that are believed to be bugs with |PRODUCT|
-   - |RsyslogStackOverflow|_ - least activity by dev team, but generally
-     good support from the wider tech community
+   - |RsyslogStackOverflow|_ - experimental support from rsyslog community
 
 .. seealso::
 
--- a/source/includes/substitution_definitions.inc.rst
+++ b/source/includes/substitution_definitions.inc.rst
@@ -49,6 +49,9 @@
 .. |RsyslogOfficialForums| replace:: Forum
 .. _RsyslogOfficialForums: http://kb.monitorware.com/rsyslog-f40.html
 
+.. |RsyslogStackOverflow| replace:: Stack Overflow
+.. _RsyslogStackOverflow: https://stackoverflow.com/questions/tagged/rsyslog
+
 .. |RsyslogOfficialMailingList| replace:: Mailing list
 .. _RsyslogOfficialMailingList: http://lists.adiscon.net/mailman/listinfo/rsyslog
 
--- a/source/includes/substitution_definitions.inc.rst
+++ b/source/includes/substitution_definitions.inc.rst
@@ -49,8 +49,11 @@
 .. |RsyslogOfficialForums| replace:: Forum
 .. _RsyslogOfficialForums: http://kb.monitorware.com/rsyslog-f40.html
 
-.. |RsyslogStackOverflow| replace:: Stack Overflow
-.. _RsyslogStackOverflow: https://stackoverflow.com/questions/tagged/rsyslog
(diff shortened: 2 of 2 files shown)


source/includes/footer.inc.rst
```diff
@@ -2,15 +2,17 @@
...
-   please review the contribution guidelines listed in the
-   |GitHubSourceProjectREADME|_.
+   Help with configuring/using |PRODUCT|:
 
-   If you have a question about these docs or |PRODUCT| in general, please
-   see the following resources:
+   - |RsyslogOfficialMailingList|_ - best route for general questions
+   - GitHub: |GitHubSourceProject|_ - best route for detailed questions and
+     for reporting issues that are believed to be bugs with |PRODUCT|
+   - |RsyslogStackOverflow|_ - least activity by dev team, but generally
```
rgerhards: I would tell that this is *experimental* -- it involves a kind of "use at your own risk, you have been told" kind of attitude IMO

rgerhards:
In general I think this looks good, but we should wait over the weekend if there is objection.

source/includes/footer.inc.rst
deoren: I can add further emphasis if you prefer/think it's necessary. I was thinking that "least activity by dev team" would cover that and would still hold true even if we end up providing more support for it over time.

deoren:
@rgerhards Alright. I didn't say it in the OP for this PR (I meant to), but I went ahead and split out "support" and "contributions" vs how it was before. With everything together it felt a little crowded.

deoren:
Some quick notes for later review (in case we do go this route):

Stack Overflow also provides a way to link to the "ask a question" form and include an initial tag:

    https://stackoverflow.com/questions/ask?tags=rsyslog

This is the link currently included in the PR (the listing of existing questions with the 'rsyslog' tag with a button off to the right to ask new questions):

    https://stackoverflow.com/questions/tagged/rsyslog

rgerhards:
> Some quick notes for later review (in case we do go this route):
> 
> Stack Overflow also provides a way to link to the "ask a question" form and include an initial tag:
> 
> https://stackoverflow.com/questions/ask?tags=rsyslog
> This is the link currently included in the PR (the listing of existing questions with the 'rsyslog' tag with a button off to the right to ask new questions):
> 
> https://stackoverflow.com/questions/tagged/rsyslog

Just followed that and think the second form (existing questions) is great. I only begun to wonder if StackOverflow is really right - or another of there myriad of web properties ;-) How do you find which is the right one, or is it visible through all the domains :-S

source/includes/footer.inc.rst
rgerhards: I would go for "experimental" as it currently looks pretty seldom that a dev will check, but it's not something I insist on...

rgerhards:
@deoren I am also talking how we can finally change the rsyslog forum to read-only state. Ping me if you don't hear back within the next 2 days. And, yes, I am approaching a state with imfile where I again begin to have "a couple of minutes" "in between (test runs)" ;-)

deoren:
> @rgerhards: I only begun to wonder if StackOverflow is really right - or another of there myriad of web properties ;-) 

I tried to craft the link to use the main stackexchange.com domain previously and it didn't work. Looking now I see an option for setting up a filter that applies to all sites. In the results I believe I see some rsyslog tagged questions from serverfault.com that do not show up in the list for stackoverflow.com.

> @rgerhards: How do you find which is the right one, or is it visible through all the domains :-S

I'll double-check to make sure that is true. If it is, then you're right: the link I proposed in the PR won't do quite what we're looking for.

After testing, I found that you can login and create a "filter" which appears to be usable for others without logging in. Optionally you can use the same filter to receive email notifications.

stackexchange_setup_filter_for_all_sites_2018-03-13

deoren:
> @deoren: After testing, I found that you can login and create a "filter" which appears to be usable for others without logging in. 

If that is the route we go, I suggest that you be the one to create it either under your account or a "service" account tied back to the rsyslog.com domain. Once you specify the parameters (e.g., as shown in the previous screenshot), we can use the new link in the docs.

deoren:
> @rgerhards: @deoren I am also talking how we can finally change the rsyslog forum to read-only state. Ping me if you don't hear back within the next 2 days. And, yes, I am approaching a state with imfile where I again begin to have "a couple of minutes" "in between (test runs)" ;-)

ACK.

rgerhards:
@deoren on StackExchange: I realized that it officially says it is targeted towards "developers" (thus I probably knew it best). So I am not sure if that really is the place admins look at... I don't think it is necessary to have an "over the network capability" -- it's more a kind of frustration on my side that they offer too many (IMHO) brandings. Up to the point that it's too fuzzy to be useful...

deoren:
> @rgerhards: I would go for "experimental" as it currently looks pretty seldom that a dev will check, but it's not something I insist on...

I'll add that in when I apply the next update to the file. I assume that we'll hold off from merging this PR until that is changed and we know for sure what URL we'll use to link over to Stack Exchange/Overflow/whatever.

> @rgerhards: it's more a kind of frustration on my side that they offer too many (IMHO) brandings. Up to the point that it's too fuzzy to be useful...

I agree. At some point if you subdivide too much you fracture your community.

deoren:
I think I missed where the forum was finally put into read-only mode. I knew it was probably coming, but I missed where it happened.

I'll review this PR tomorrow and update it to reflect the recent conversations that we've had on this. I'll ping you when it's ready for review.

rgerhards:
Maybe I forgot to notify you. Sorry if that's the case.

deoren:
> @rgerhards: Maybe I forgot to notify you. Sorry if that's the case.

Not a problem, I wasn't trying to blame you, just noting that I've been busy with other stuff and missed it.

deoren:
@rgerhards If you haven't yet, see  https://github.com/rsyslog/rsyslog-doc/pull/624#issuecomment-372731827.

Did you want to create such a filter for our use here? In case I didn't do a good job explaining it there, the idea is to create a "filter" (accessible by others without logging in) to create a view across all Stack Exchange properties for the `rsyslog` tag. Otherwise, as you noted it appears we would have to create multiple links to encompass them all.

rgerhards:
I don't necessarily think that we need an across-all-sites filter. My main concerns is that stack overflow is actually the wrong subsite, see https://stackexchange.com/sites - for StackOverflow it gives the target audience as "Q&A for professional and enthusiast programmers" while for ServerFault it gives "Q&A for system and network administrators". From that I would say we target the wrong site.

deoren:
> @rgerhards: I don't necessarily think that we need an across-all-sites filter. My main concerns is that stack overflow is actually the wrong subsite, see https://stackexchange.com/sites - for StackOverflow it gives the target audience as "Q&A for professional and enthusiast programmers" while for ServerFault it gives "Q&A for system and network administrators". From that I would say we target the wrong site.

I get what you're saying in terms of "purpose", but if you look where users are actually posting their questions I don't think they're paying as much attention to the purpose of each site as you are (which is unfortunate). I see the most 'rsyslog' tagged posts in the filtered list coming from stackoverflow.com with a spread of other properties mixed in.

Here is a sampling:

- stackoverflow.com (seems to have the most questions with that tag)
- serverfault.com
- superuser.com
- unixstackexchange.com

I created a filter and sent it to you by email to give you an idea of the results.

I'm not saying the filter is the ideal approach, but it looks like a workable one based on where users are actually posting.

rgerhards:
Well, given that mess (no wonder @100+ sites...), I would actually find any solution good to give a try. So go ahead with whatever you find useful - you are closer to the target user base than I am.

deoren:
> @rgerhards: Well, given that mess (no wonder @100+ sites...), I would actually find any solution good to give a try. So go ahead with whatever you find useful - you are closer to the target user base than I am.

What do you think as of the changes in afd21ef41d70cbc3c81d06d7742dd889d6c11870?

deoren:
@rgerhards Thanks for the feedback.

The only potential "gotcha" here is the filter I setup for viewing the 'rsyslog' tagged posts against all Stack Exchange properties. I _assume_ that the filter link will keep working if my account is closed, but I don't know that for sure. It may be worth recreating the filter under your account or an adiscon.com account set aside specifically for the purpose.

If you're alright with the risk, please merge this when ready. If you believe it's worth setting up the filter using another account (one owned by adiscon.com), let me know if there is anything I can do to help with that and I'll do so. We can then replace the existing filter link currently in the substitution definitions doc with the new link.

rgerhards:
I think we can bear the risk 😉

deoren:
> @rgerhards: I think we can bear the risk

:)

Issue #467, opened 2018-01-01
Still able to find old(?) pages


jgerhards:
there are some old (?) pages that can still be reached that are not up to date and do not appear to be maintained. Some quick examples I found include:
http://www.rsyslog.com/doc/mmjsonparse.html
http://www.rsyslog.com/doc/mmanon.html

deoren:
@alorbach @friedl It looks like the /doc/ directory contains content not generated from the reST source files in this repo. Can you confirm that generated content from this repo goes into either the `v8-stable` or the `master` directories on rsyslog.com, and not into the /doc/ path on rsyslog.com?

rgerhards:
@deoren sry, I should have commented on this. I talked with @friedl actually past monday, and it is most probably caused by redirect rules which I created to keep URLs stable. I need to review if that is still needed (gut feeling: no). Hope to have time over the weekend to look at it. I wouldn't mind if you ping me if not done by mid next week ;-)

deoren:
@rgerhards Thanks, will do.

deoren:
@rgerhards A little late, but here is the requested "ping".

Issue #420, opened 2017-12-11
Review v1 files, prune/merge into current docs as necessary


deoren:
From https://github.com/rsyslog/rsyslog-doc/pull/416#issuecomment-350542356:

> @rgerhards: I forgot to mention that I see a lot of v1 files which we probably want to get rid of sooner or later. Just FYI...

Have those v1 files been superseded by newer files? In the past I landed on some of them when searching via Google and found the same content on newer pages. I don't recall if that was the case for all of the older v1 files.

rgerhards:
> Have those v1 files been superseded by newer files? In the past I landed on some of them when searching via Google and found the same content on newer pages. I don't recall if that was the case for all of the older v1 files.

actually (I think, validation needed to be 100% sure) these are description of obsolete format, which we wanted to refer to v5 rather than including. So I would expect that most of the files can simply be deleted (or put on a summary page). But that's a hard question, I think it's better to answer once we have moved forward a bit more.

deoren:
#625 is likely related to this

PR #625, opened 2018-03-11
fix typo in ActionResumeInterval description

nohararc:


deoren:
@nohararc Thank you for taking the time to craft this PR! I took a look at the doc files that were modified and compared against the current source code to make sure everything lined up.

Here is what I found.

`~/dev/rsyslog/rsyslog$ grep -ri "ActionResumeInterval" .`

```
./tests/testsuites/rcvr_fail_restore_sender.conf:$ActionResumeInterval 1
./ChangeLog:- added config file directive $ActionResumeInterval
./ChangeLog:- added config file directive $ActionResumeInterval
./ChangeLog:- added config file directive $ActionResumeInterval
./runtime/rsconf.c:static rsRetVal setActionResumeInterval(void __attribute__((unused)) *pVal, int iNewVal)
./runtime/rsconf.c:     CHKiRet(regCfSysLineHdlr((uchar *)"actionresumeinterval", 0, eCmdHdlrInt,
./runtime/rsconf.c:             setActionResumeInterval, NULL, NULL));
./action.c:     int glbliActionResumeInterval;
./action.c:     cs.glbliActionResumeInterval = iNewVal;
./action.c:             pAction->iResumeInterval = cs.glbliActionResumeInterval;
./action.c:     cs.glbliActionResumeInterval = 30;
```

This confirms that the default value is 30 as you've indicated.

Unfortunately we're dealing with a combination of problems:

- The `source/configuration/action/rsconf1_actionresumeinterval.rst` file is for some older documentation that has been deprecated, but not removed yet (refs #420).

Link to deprecated file: http://www.rsyslog.com/doc/rsconf1_actionresumeinterval.html (refs #467). 

The good aspect of the older document vs the newer copy is that the older makes it very clear what the default value is. For the newer doc (http://www.rsyslog.com/doc/master/configuration/actions.html), this is what is currently noted for the parameter:

> Sets the ActionResumeInterval for the action. The interval provided is always in seconds. Thus, multiply by 60 if you need minutes and 3,600 if you need hours (not recommended). When an action is suspended (e.g. destination can not be connected), the action is resumed for the configured interval. Thereafter, it is retried. If multiple retries fail, the interval is automatically extended. This is to prevent excessive resource use for retries. After each 10 retries, the interval is extended by itself. To be precise, the actual interval is (numRetries / 10 + 1) * Action.ResumeInterval. so after the 10th try, it by default is 60 and after the 100th try it is 330.

I think the confusion comes from this sentence:

> To be precise, the actual interval is (numRetries / 10 + 1) * Action.ResumeInterval. so after the 10th try, it by default is 60 and after the 100th try it is 330.

In particular, the **it by default is 60** phrase.

Instead, it should probably be worded something similar to this:

> To be precise, the actual interval is (numRetries / 10 + 1) * Action.ResumeInterval. Using the default value of `30`, this means that on the 10th try the suspension interval will be **60** (seconds) and after the 100th try it will be **330** (seconds).

In short I think the default is correct as was the original math, but the wording is confusing. Please double-check me on that to make sure I didn't mix something up, but I think I'm right.

On a related note, we'll likely be updating this doc at some point to reflect a newer format that attempts to make the default values for the parameters much easier to find.

Here is a sample of how the values will be displayed:

http://www.rsyslog.com/doc/master/configuration/modules/ommysql.html

Feedback is very welcome.

Thanks again for your interest in contributing. There is a ton of work to be done and every bit helps!

nohararc:
@deoren Thanks!

deoren:
@nohararc Sure. My apologies if it seemed like I was rejecting your PR, instead I was trying to frame some changes to the docs that we could apply together. Please do submit PRs for any further improvements to the docs and we'll be happy to work with you to apply them.

deoren:
Related note: I submitted PR #629 with the suggested changes that I noted here. It's not the complete overhaul that is needed, but it's a quick fix for the time being. Please feel free to point out anything else you spot out of place. We can use all the help we can get.

PR #629, opened 2018-03-12
Clarify how action.resumeInterval increases over time

deoren:
Attempt to clarify the language for how the interval increases over time based off of feedback from @nohararc on rsyslog/rsyslog-doc#625.

Further work needs to be done to make clear the default value for all parameters, but that is outside the scope of this change.


rgerhards:
LGTM

PR #630, opened 2018-03-20
Link directly to properties documentation

cincodenada:
To see a list of properties, one has to follow the link from the Property Replacement page to get to the "rsyslog Properties" page. Might as well just link directly there from here. I imagine this may have happened becasue the rsyslog Properties didn't used to be a separate page?

### DIFF
--- a/source/configuration/filters.rst
+++ b/source/configuration/filters.rst
@@ -76,11 +76,9 @@ Property-Based Filters
 
 Property-based filters are unique to rsyslogd. They allow to filter on
 any property, like HOSTNAME, syslogtag and msg. A list of all
-currently-supported properties can be found in the :doc:`property replacer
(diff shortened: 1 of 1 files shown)


deoren:
@cincodenada Thank you for taking the time to craft and submit a PR. 

> @cincodenada: To see a list of properties, one has to follow the link from the Property Replacement page to get to the "rsyslog Properties" page. Might as well just link directly there from here.

Good catch!

> @cincodenada: I imagine this may have happened becasue the rsyslog Properties didn't used to be a separate page?

It wouldn't surprise me.

rgerhards:
> @cincodenada: I imagine this may have happened becasue the rsyslog Properties didn't used to be a separate page?
> 
> It wouldn't surprise me.

That's exactly how it happened!

PR #632, opened 2018-03-23
fixed a typing error.

rupav:
2 'automating' were used together, removed 1 .


deoren:
@rupav Thank you for taking the time to craft and submit this PR! Thanks also for fixing one of my mistakes. :)

PR #633, opened 2018-03-23
add documentation for the change to module and built-in functions

jgerhards:
also add description of http_request function

### DIFF
--- a/source/rainerscript/functions.rst
+++ b/source/rainerscript/functions.rst
@@ -1,8 +1,27 @@
 Functions
 =========
 
-RainerScript supports a currently quite limited set of functions:
+There are two types of RainerScript functions: built-ins and modules. Built-in
+functions can always be called in the confiuration. To use module functions,
+you have to load the corresponding module first. To do this, add the following
+line to your configuration:
 
+::
+
+	module(load="<name of module>")
+
+If more than one function with the same name is present, the function of the
+first module loaded will be used. Also, an error message stating this will be
+generated. However, the configuration will not abort. Please note that built-in
+functions will always be automatically loaded before any modules. because of this, you
+are unable to override any of the built-in functions, since their names are already
+in use. The name of a function module starts with fm.
+
+
+built-in functions
+==================
+
+You do not have to load any module to use these functions.
 
 getenv(str)
 -----------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -145,6 +145,8 @@ later steps are identical, so those steps have been covered in one place.
 
 #### Windows and Linux
 
+1. Install python-pip
+    1. `sudo apt install python-pip`
```
deoren: See the "Linux" section earlier in this document. There we install pip via `python ./get-pip.py --user` (using the previously downloaded script from upstream) and then setup a virtualenv. If you believe that installing via package is better, it may be worth moving this step specifically to the "Linux" section and then updating the rest of the doc to replace the earlier references to installing via the `get-pip.py` script. I don't recall how many there are, there might just be the one.

jgerhards:
Yes, forgot to mention it.

deoren:
@jgerhards No rush on responding, but I wanted to add a quick note here on the main issue that I made an in-line review comment. Basically it looks like earlier directions already provide a step to make `pip` available for the user. Installing via package is fine, but we should probably update the earlier directions to omit installing via the upstream installation script (which I think was used here to cover both Linux and Windows build platforms).

jgerhards:
tbh I'm not really sure it is better this way, but in the document it seemed overly complicated to set up the doc. Because of this (and me remembering there was a faster way to setup) I asked @rgerhards and he provided the direction i added. Not sure though, only setup the doc twice (counting this time) but have some experience installing packages.

rgerhards:
sorry for the short comment, but thought I at least add this: I have reviewed README and IMHO at has grown to a size that can make one feel uncomfortable for doing a quick change. Also, many non-pythonish folks may prefer to do a quick package install vs a (only seemingly more complicated) "real install procedure". I admit that I also prefer to use packages for non-essential things I do because I have too often seen side effects of installing other ways. I know it's a mood argument, and I know that rsyslog itself is notoriously bad at this easy setup, but I still think it's a point to consider. I do not know, unfortunately, how to address that decently. Maybe shrink README to an absolute minimum and put the rest into an other file? But that also has its downsides...

rgerhards:
thx for removing the README change. @deoren we still need to think about that point, but let's do that separately.

Merging now, as the underlying code had already to be merged due to other needs.

deoren:
> @rgerhards: @deoren we still need to think about that point, but let's do that separately.

ACK.

davidelang:
I'll go on record saying that I don't really like the idea of grabbing the 
'latest on the internet' as opposed to a version packaged with the distro. I've 
been bit a few times over the years by the 'latest from the internet' instead of 
'the version I used last time'

There are times you need to do this, but I try to avoid this as a general rule.

PR #635, opened 2018-03-25
"basic" format doc tweaks: formatting, headers

deoren:
- adjust header levels
- add newline near 80 char mark
- code formatting on some keywords, filenames (more to do here)
- add Sphinx cross-reference manpage role to manpage refs

There is more to do, but I wanted to go ahead and apply a quick pass now while my attention was still on the doc (as part of working with another PR).

### DIFF
--- a/source/configuration/sysklogd_format.rst
+++ b/source/configuration/sysklogd_format.rst
@@ -1,282 +1,315 @@
+
+***************
 sysklogd format
-~~~~~~~~~~~~~~~
+***************
 
 This is the format in use since the beginning of syslogging. It still
 is an excellent choice to do very simple things.
 
-For more advanced things, use RainerScript format.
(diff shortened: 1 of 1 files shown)


rgerhards:
I'd say lets merge. Just be warned that I considered this page as a "let's fix it quickly and dirty" shot. My intent was to remove it again as most of the content is also in the rest of the doc set or at least should be there. In that line of thought, the page should probably just become links to dedicated sections. To stay, I think it also simply is too long and needs to be split in separate pages - but IMHO more important things to do at the moment than digging down on this one. And so improvements are definitely nice, even if they only tactically help for a couple of weeks (what often enough becomes months and years...).

deoren:
@rgerhards ACK.

As you noted, some quick polish until the larger aspect of the work can be tackled.

PR #640, opened 2018-04-04
Adding documentation for omkafka impstats counters

abwaheed:
This is the doc related to PR: 
https://github.com/rsyslog/rsyslog/pull/2598

### DIFF
--- a/source/configuration/modules/omkafka.rst
+++ b/source/configuration/modules/omkafka.rst
@@ -300,6 +300,86 @@ failedMsgFile
 Filename where the failed messages should be stored into.
 Needs to be set when keepFailedMessages is enabled, otherwise failed messages won't be saved.
 
+Statistic Counter
+=================
+
+This plugin maintains global :doc:`statistics <../rsyslog_statistic_counter>` for omkafka that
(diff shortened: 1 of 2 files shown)


deoren:
@abwaheed Thanks for submitting this PR. I don't know enough about the module to provide any feedback, but the general format of the doc changes looks good. I would be tempted to break the bullet points into subsections, but I don't know that the extra depth is needed here and might make it harder to see all of the counters at a glance.

rgerhards:
we need to squash, can do so during merge

rgerhards:
@deoren 
> but I don't know that the extra depth is needed here and might make it harder to see all of the counters at a glance.

I think it is a great as suggested. The ultimate reference is the librdkafka link quoted (this is where the counters originate from). So a short list that gives a rough idea is exactly right IMHO.

deoren:
>> @deoren: but I don't know that the extra depth is needed here and might make it harder to see all of the counters at a glance.
> 
> @rgerhards: I think it is a great as suggested. The ultimate reference is the librdkafka link quoted inside the doc PR (this is where the counters originate from). So a short list that gives a rough idea is exactly right IMHO.

ACK, sounds like we're on the same page.

PR #637, opened 2018-03-27
rainerscript: function documentation split up

PascalWithopf:
The page for rainerscript functions was split up into multiple pages to make it easier to read.

closes https://github.com/rsyslog/rsyslog-doc/issues/634

PR #641, opened 2018-04-04
rainerscript functions: add information

PascalWithopf:
Improvement ideas of following comment were added.
https://github.com/rsyslog/rsyslog-doc/pull/637#issuecomment-376973382

### DIFF
--- a/source/rainerscript/functions/idx_built-in_functions.rst
+++ b/source/rainerscript/functions/idx_built-in_functions.rst
@@ -8,4 +8,8 @@ Following functions are available in Rainerscript.
    :glob:
    :maxdepth: 1
 
+   ipv42num() <rs-ipv4convert.rst>
+   num2ipv4() <rs-ipv4convert.rst>
+   ltrim() <rs-trim.rst>
+   rtrim() <rs-trim.rst>
    rs*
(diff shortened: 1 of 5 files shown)


PascalWithopf:
@deoren @davidelang ltrim/rtrim and ipv42num/num2ipv4 have now separate lines on the main page and new pages will still be added automatically, but it is not entirely alphabetic now.

There is no toupper() function which I'm aware of.

deoren:
> @PascalWithopf: @deoren @davidelang ltrim/rtrim and ipv42num/num2ipv4 have now separate lines on the main page and new pages will still be added automatically

That's pretty clever. Because you've already explicitly listed them, Sphinx won't duplicate them in the wildcard match used a few lines down.

> @PascalWithopf: but it is not entirely alphabetic now.

A small problem, one perhaps we can solve later on as a solution becomes more obvious. If it were particularly important to keep the alphabetical order, we could break the combined pages apart and use include directive calls to stitch everything together. Not sure it's worth the effort though.

Setting @davidelang as the reviewer for this based on his extensive feedback on the last PR. I've only given this a brief review, so it's likely that @davidelang will catch something I've missed.

davidelang:
looks good to me

PR #645, opened 2018-04-10
DO NOT MERGE - just trigger a CI build based on current master

rgerhards:
**DO NOT MERGE** - *discard when done*

test trying to find out what's going wrong in #558

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 sudo: required
+group: deprecated-2017Q4
 services: docker
 language: minimal
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,12 @@ install:
   - docker pull rsyslog/rsyslog_doc_gen
 
 script:
+  - git log -4
+  - git --version
+  - git describe ; true
+  - git describe HEAD~1 ; true
+  - git describe HEAD~2 ; true
+  - git describe HEAD~3 ; true
   - docker run -e SPHINX_EXTRA_OPTS=-q
         -v "$PWD":/rsyslog-doc -ti
         rsyslog/rsyslog_doc_gen
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+REMOVE THIS LINE - TEST ONLY
 # rsyslog-docs
 
 ## Documentation for the rsyslog project
--- a/source/conf_helpers.py
+++ b/source/conf_helpers.py
@@ -78,12 +78,7 @@ def get_next_stable_version():
 def get_current_commit_hash():
     """Return commit hash string"""
 
-    # e.g., v8.29.0-98-g07d02c6 (after decoding)
-    # https://stackoverflow.com/questions/2502833/store-output-of-subprocess-popen-call-in-a-string
-    #git_describe_output = subprocess.check_output(['git', 'describe']).decode("utf-8")
-
-    # Grab the last portion of the strings, then strip off leading 'g'
-    commit_hash = "master" # git_describe_output.split('-')[-1][1:].strip()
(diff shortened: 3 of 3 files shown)


deoren:
This is odd. What happens when building the docs from master locally using the same Docker container?

rgerhards:
> This is odd. What happens when building the docs from master locally using the same Docker container?

This would not be a valid test, as they would have different commits. I added ```git log``` to Travis. It shows:
```
Merge: 33a2791 0591e5f
Author: Rainer Gerhards <rgerhards@adiscon.com>
Date:   Tue Apr 10 13:53:21 2018 +0000
    Merge 0591e5ff0000d2bef02b70f30656f3e9dcc4e352 into 3a2791465ae5bed427988ef1e2f1f336c78a0fe

commit 0591e5ff0000d2bef02b70f30656f3e9dcc4e352
Author: Rainer Gerhards <rgerhards@adiscon.com>
Date:   Tue Apr 10 15:50:02 2018 +0200
    do not query commit hash in conf.py
```
See the new merge commit on top of mine? This is done when Travis runs. Possibly something has changed inside Travis (git version?) in how exactly this is done.

deoren:
> @rgerhards: This would not be a valid test, as they would have different commits.

Sounds like that is true based on the behavior you observed, but I thought previously that Travis just did a shallow checkout of the branch used by the PR and built the docs as-is. Of course, "how it was done before" != "how it is done now".

rgerhards:
> Sounds like that is true based on the behavior you observed, but I thought previously that Travis just did a shallow checkout of the branch used by the PR and built the docs as-is. Of course, "how it was done before" != "how it is done now".

no, it merges

rgerhards:
from https://docs.travis-ci.com/user/pull-requests/:
```
Rather than build the commits that have been pushed to the branch the pull request is from, we build the merge between the source branch and the upstream branch.
```

deoren:
> from https://docs.travis-ci.com/user/pull-requests/:
> 
> Rather than build the commits that have been pushed to the branch the pull request is from, we build the merge between the source branch and the upstream branch.

https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/pull-requests.md

Looks like that file was last modified two months ago too, so definitely not a sudden change in behavior on that point (though it sounds like you have personal experience with how the build process is initiated regardless of the modification date of the Travis CI doc page).

So much for that thought.

rgerhards:
ok, whatever the root cause is, the problem is the ```git describe```. I have commented it out, and now everything works. As we can see, ```git describe``` also fails via the command line.

rgerhards:
note: git on Travis is 2.15.1. I have up to 2.14.1 immediately at hand, and it works on the same git log (note that not only the merge commit fails!)

rgerhards:
Thinking about it, it can't be the git version -- the container is unchanged. So something with the travis checkout smells incorrect (or at least changed).

rgerhards:
Falling back to the previous travis image via .travis.yml
```
group: deprecated-2017Q4
```
does not help. Probably time for a support call at Travis.

rgerhards:
Let's first try getting the commit hash directly (w/o ```git describe```). The actual command (**for Travis!**) is:
```
git log --pretty=format:'%h' HEAD~1 -n1
```

rgerhards:
This seemed to work for Travis. However, now everywhere else we get the previous commit hash ;-) Thinking about it "everywhere else" is really what we are after, so using HEAD commit in Travis as well, which is wrong ... but nobody ever knows that. So looks like a solution.

rgerhards:
#646 has a solution

rgerhards:
support request to Travis sent. Will report the outcome. For this, I leave this test PR here open for a moment...

PR #646, opened 2018-04-10
use different method to obtain commit hash

rgerhards:
Most importantly, Travis CI stopped working with the old method
on 2018-04-10. In any case, the new method looks more up to the
point

### DIFF
--- a/source/conf_helpers.py
+++ b/source/conf_helpers.py
@@ -78,12 +78,7 @@ def get_next_stable_version():
 def get_current_commit_hash():
     """Return commit hash string"""
 
-    # e.g., v8.29.0-98-g07d02c6 (after decoding)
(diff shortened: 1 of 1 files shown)


rgerhards:
@deoren any concerns merging this? Seems to nail #645 and generally looks simpler to me.

deoren:
> @rgerhards: @deoren any concerns merging this? Seems to nail #645 and generally looks simpler to me.

Simpler works for me. As you say, this is much more direct re the intention. I wasn't (and still am not) that fluent with Git, so I wasn't aware there was a better way. Thanks for looping me in so I could learn a better way. :)

rgerhards:
@deoren I think I found this method when I did the buildbot CI integration, where I need the hash for the directory name. Didn't think about the python part at that time. But now I remembered ;-)

PR #647, opened 2018-04-12
mmkubernetes doc updates to go with recent fixes

richm:


richm:
goes with https://github.com/rsyslog/rsyslog/pull/2635

PR #657, opened 2018-04-27
omprog: rewrite documentation

PascalWithopf:
omprog documentation was rewritten in the new style.

### DIFF
--- a/source/configuration/modules/omprog.rst
+++ b/source/configuration/modules/omprog.rst
@@ -1,13 +1,15 @@
+*****************************************
 omprog: Program integration Output module
-=========================================
+*****************************************
 
-**Module Name:    omprog**
+===========================  ===========================================================================
+**Module Name:**             **omprog**
+**Author:**                  `Rainer Gerhards <http://rainer.gerhards.net/>`_ <rgerhards@adiscon.com>
+===========================  ===========================================================================
 
-**Available since**: 4.3.0
 
-**Author:**\ Rainer Gerhards <rgerhards@adiscon.com>
-
-**Description**:
+Purpose
+=======
 
 This module permits to integrate arbitrary external programs into
 rsyslog's logging. It is similar to the "execute program (^)" action,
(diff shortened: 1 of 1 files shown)


PascalWithopf:
@deoren While working on this pull request I noticed that only part of the parameter are documented and I am trying to figure out why at the moment.
What I can say is that the four parameter with upper case letters in their name don't work because the names need to be lower case. They can not be used.
The other two parameter
```
{ "output", eCmdHdlrString, 0 },
{ "forcesingleinstance", eCmdHdlrBinary, 0 },
```
can be used but weren't documented.

rgerhards:
@jsiwrk something looks fishy here... if the params are given not in lower case, they can never be selected. However, I think the testbench makes use of them. I am puzzled - do you have an explanation?

jsiwrk:
Yes, I plan to add the missing documentation for the new parameters. I priorized the tests so far.

Since I will have to learn the documentation format, I suggest accepting this PR as is (rather than deferring it to include the new params), since this update to the new documentation style will make things easier for me, I imagine.

Regarding the case: I didn't know this rule. I suppose it's been working for me all the time (and in the tests) because I always write the parameters in the exact case they are declared (?). I will check this and correct the declarations to lowercase in another PR.

rgerhards:
> Regarding the case: I didn't know this rule. I suppose it's been working for me all the time (and in the tests) because I always write the parameters in the exact case they are declared (?). 

This puzzles me a bit, because it cannot work: rsyslog converts all parameter names to lower case and then does an exact comparison. I think it's a left-over of the optimizations inside the core: strcasecmp is considerably slower than strcmp. In this context (config reading) speed doesn't really matter, so we should probably drop that requirement. Nevertheless, these parameters never worked, as so if they are essential for the tests, I wonder why they worked...

jsiwrk:
I don't know the details, but I can ensure that these parameters are working.

rgerhards:
oops... sorry for the confusion: looks like I already changed that to strcasecmp... Had forgotten that :-(

jsiwrk:
@deoren As I said, I suggest merging this PR even though the documentation of the recently added params is still missing. I promise I will add these in a future PR, when I can find some time...

davidelang:
can you at least put placeholders in there to say that the parameters exist?

deoren:
>@jsiwrk: @deoren As I said, I suggest merging this PR even though the documentation of the recently added params is still missing. I promise I will add these in a future PR, when I can find some time...

Acknowledging your remark. Please do not take my silence as ignoring your feedback, I've just been really busy of late and haven't been able to dedicate as much time to doc efforts as I have in the past.

> @davidelang: can you at least put placeholders in there to say that the parameters exist?

This is a good suggestion.

@PascalWithopf Are you up for stubbing in the other parameters as part of this PR or do you feel that exceeds the original scope of this work? 

If the latter, perhaps as a compromise go ahead and create a new issue for adding the remaining parameters and then we can merge this PR as-is.

jsiwrk:
I think I will be able to submit a PR documenting the new params within about two weeks. May be it is not worth stubbing them...

rgerhards:
@PascalWithopf is this done? Looks like it is so, but I admit I am too lazy right now to check the full thread and parameter names ;-)

PascalWithopf:
@rgerhards from my point of view this is done. I have added placeholders for all other parameters.

PR #660, opened 2018-04-28
describe new global parameter "internal.developeronly.options"

rgerhards:


deoren:
related to rsyslog/rsyslog#2675?

PR #661, opened 2018-04-30
omruleset: rewrite documentation

PascalWithopf:
documentation was rewritten in the new style


rgerhards:
Not sure if the current wording is sufficient: this module is horrible outdated and only provided to support configurations that used it (before we had the ```call``` statement). NOT to be used for anything new.

PascalWithopf:
@rgerhards I added a warning at the beginning. I think this is sufficient.

PR #666, opened 2018-05-14
Add example to dateformat and link to property options

PascalWithopf:
To make the use of dateformat clearer a link to the
options and an example.

Also the format was changed to a newer style


rgerhards:
Note: I intend to merge this after only quick review, as I would like to have it in 8.35 release. @deoren should you see further issues, please let us work on them later on.

rgerhards:
@deoren added you as reviewer as a hint that a review is appreciated even after merge.

deoren:
> @rgerhards: @deoren added you as reviewer as a hint that a review is appreciated even after merge.

ACK. I'll try to work this in sometime this week.

rgerhards:
That's a good question. Normally I would not have merged, so it is an
exception... Let's see what @pascalwithopf means and does.

Rrainer

PR #678, opened 2018-05-28
omprog: document missing params, general review

jsiwrk:
Add documentation for confirmMessages, useTransactions (and related),
closeTimeout, killUnresponsive.

signalOnClose kill behaviour no longer linux-specific.

Add more complete example illustrating the feedback feature.

forceSingleInstance should be considered deprecated.

### DIFF
--- a/source/configuration/modules/omprog.rst
+++ b/source/configuration/modules/omprog.rst
@@ -50,7 +50,7 @@ fork. As the actual execution must happen after the fork, we cannot
 use the default error logger to emit the error message. As such,
 we use ``syslog()``. In most cases, there is no real difference
 between both methods. However, if you run multiple rsyslog instances,
-the messae shows up in that instance that processes the default
+the message shows up in that instance that processes the default
 log socket, which may be different from the one where the error occured.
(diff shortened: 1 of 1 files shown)


jsiwrk:
Please see [this comment](https://github.com/rsyslog/rsyslog/issues/473#issuecomment-392350337) for the reason I'm proposing to document the `forceSingleInstance` flag as deprecated.

jsiwrk:
@rgerhards : Do you agree then with my analysis of the forceSingleInstance flag, and with documenting it as deprecated?

If you agree, I can submit a PR for omprog documenting the flag as deprecated in the code. Or at least explaining what it really does (since the name is misleading).

rgerhards:
I have commented on the quoted issue tracker - it's actually a bug in omprog, the flag should work as described. Good catch!

jsiwrk:
In this PR I documented the `forceSingleInstance` flag as deprecated, for the reasons explained above. However, this could be wrong: see issue #2813. I will submit another documentation PR amending this if necessary.

PR #693, opened 2018-06-14
correct link to rainer.gerhards.net

PascalWithopf:


rgerhards:
failures are due to v7-stable. @PascalWithopf looks like a small issue, can you fix it with a separate commit? Then we do not need to think about this each time we need to update the old stuff...

PascalWithopf:
Upon looking into the issue we noticed that the issue is not so small. We decided to leave it without further changes.

PR #694, opened 2018-06-14
tutorial: change example to new config style

PascalWithopf:


source/tutorials/failover_syslog_server.rst
```diff
@@ -12,16 +14,19 @@ Let's assume you have a primary and two secondary central servers. Then, you
 can use the following config file excerpt to send data to them:
 rsyslog.conf:
 
-::
+.. code-block:: none
+
+   action(type="omfwd" target="primary-syslog.example.com" port="15314"
+                protocol="tcp")
+   & action(type="omfwd" target="secondary-1-syslog.example.com" port="15314"
```
rgerhards: The & is no longer needed and may even cause harm (not 100 sure) - did you run the config once through rsyslog?

source/tutorials/failover_syslog_server.rst
```diff
@@ -12,16 +14,19 @@ Let's assume you have a primary and two secondary central servers. Then, you
...
 
-  *.* @@primary-syslog.example.com
-  $ActionExecOnlyWhenPreviousIsSuspended on
-  & @@secondary-1-syslog.example.com
-  & @@secondary-2-syslog.example.com
-  & /var/log/localbuffer
-  $ActionExecOnlyWhenPreviousIsSuspended off
 
-This selector processes all messages it receives (*.*). It tries to forward 
+This selector processes all messages it receives. It tries to forward
```
rgerhards: there is no longer a selector present ;-)

To make things a bit clearer, I would add an if condition, so folks can see the block-like structure, e.g.
```
if($msg contains "sometext) {
    action()
    action()
    ...
}
```

source/tutorials/failover_syslog_server.rst
```diff
@@ -12,16 +14,21 @@ Let's assume you have a primary and two secondary central servers. Then, you
...
+.. code-block:: none
+
+   if($msg contains "error") then {
+        action(type="omfwd" target="primary-syslog.example.com" port="15314"
+               protocol="tcp")
+        action(type="omfwd" target="secondary-1-syslog.example.com" port="15314"
+               action.execOnlyWhenPreviousIsSuspended="on")
+        action(type="omfwd" target="secondary-2-syslog.example.com" port="15314"
+               action.execOnlyWhenPreviousIsSuspended="on")
+        action(type="imfile" tag="failover" file="/var/log/localbuffer"
```
rgerhards: I guess this should be omfile...

PR #705, opened 2018-07-06
fix a typo in a couple documents

infrastation:


infrastation:
One more commit, sorry, I am still reading the documentation to find what I need...

rgerhards:
no problem at all, these fixes are very helpful. I'll keep the PR a bit more open. Feel free to push as you see fit!

infrastation:
I am done with my original  task and all fixups I could identify are in this pull request.

JFYI, I was looking to filter out a specific message from a specific process, because it is a binary that comes in an OS package and due to a packaging error the new version spams syslog with a specific debug message on every server. To avoid rebuilding the package company-wide I was looking to put a configuration file into `/etc/rsyslog.d` that would match on a combination of properties and delete exactly that debug message. I was expecting something like this to be available:
```
:syslogfacility-text, equals, "daemon" && :syslogseverity-text, equals, "info" && :programname, equals, "snmpd" && :msg, contains, "Connection from UDP" stop
```
It turns out it is impossible to match more than one property in a rule, so I ended up with a rule using `regex` on `rawmsg`, which worked.

davidelang:
you just can't use that syntax

you can say

if $syslogfacility-text == "daemon" and syslog severity-text == "info" and 
$programname == "snmpd" and $msg contains "Connection from UDP" then stop

infrastation:
Thank you, I will give this a try. That's RainerScript and the "Filter Conditions" documentation page describes it as a work in progress, so I didn't look further.

infrastation:
After minor syntax fixups the RainerScript filter works too, thank you! The pull request will remain open though.

rgerhards:
If you remember the place where it says "RainerScript is work in progress", it would be good to fix that, too.

infrastation:
The page https://www.rsyslog.com/doc/v8-stable/configuration/filters.html does not literally say "work in progress", but it explains:
> Expression filters will evolve into a full configuration scripting language. Unfortunately, their syntax will slightly change during that process. So if you use them now, you need to be prepared to change your configuration files some time later. However, we try to implement the scripting facility as soon as possible (also in respect to stage work needed). So the window of exposure is probably not too long.

Which originally made me think it was still in development, especially that in my case the rsyslog OS packages are a few releases behind the latest stable version.

rgerhards:
thx, that's a left-over and can completely be removed.

PR #708, opened 2018-07-19
omprog: document issues with forceSingleInstance and output redirection

jsiwrk:
The forceSingleInstance flag is currently not working (see rsyslog/rsyslog#2813).
Added a warning explaining this, along with a workaround.

The stdout/stderr of the external program are currently assigned in an
unpredictable way (see rsyslog/rsyslog#2787). Added a warning explaining this.

### DIFF
--- a/source/configuration/modules/omprog.rst
+++ b/source/configuration/modules/omprog.rst
@@ -25,13 +25,6 @@ terminates, the program's stdin will see EOF. The program must then
 terminate. The message format passed to the program can, as usual, be
 modified by defining rsyslog templates.
 
-Note that each time an omprog action is defined, the corresponding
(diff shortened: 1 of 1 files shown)


jsiwrk:
I have also removed a paragraph from the introductory section which seemed unnecessary and could cause confusion with the explanation now given in the forceSingleInstance parameter about how many program instances are started. @rgerhards, do you agree?

rgerhards:
> I have also removed a paragraph from the introductory section which seemed unnecessary and could cause confusion with the explanation now given in the forceSingleInstance parameter about how many program instances are started. @rgerhards, do you agree?

sry for the late reply - yes, I think that was good to do!

PR #741, opened 2018-11-06
Update tls.rst

mcalr3:
Propose to update the client documentation to include the certificate. I was working on getting client rsyslog set up for over a week and not until I did detailed GnuTLS debugging on the destination I found that it would not accept my TLS syslogs unless I also sent the cert along with the client.

### DIFF
--- a/source/tutorials/tls.rst
+++ b/source/tutorials/tls.rst
@@ -145,13 +145,14 @@ Client Setup
 ~~~~~~~~~~~~
 
 The client setup is equally simple. You need less certificates, just the
-CA cert. 
+CA & client cert. 
 
     ::
 
-        # certificate files - just CA for a client
+        # certificate files - just CA & cert for a client
         $DefaultNetstreamDriverCAFile /path/to/contrib/gnutls/ca.pem
-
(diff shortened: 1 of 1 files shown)


rgerhards:
@alorbach do we actually need the cert if running in anon mode (at least on ossl if you do not know for gtls)?

alorbach:
Not needed in anon mode, we are using ciphers that support certless communication.

alorbach:
There does not seem to be test. Perhaps it isn't working?

source/tutorials/tls.rst
```diff
@@ -145,13 +145,14 @@ Client Setup
 ~~~~~~~~~~~~
 
 The client setup is equally simple. You need less certificates, just the
-CA cert. 
+CA & client cert.
```
rgerhards: We need to change this so that folks know they need the client cert only if the actually want authentication (and not anon mode).

rgerhards:
@alorbach would be great if we could add a test to rsyslog testbench to ensure anon works as expected

rgerhards:
This has been changed in the interim and now fully works as documented.

Issue #743, opened 2018-11-13
Git commit quoted in docs title - is this intentional?


deoren:
Looks like with 3cde0e1b21468b74feb1a640af8613219f2af5f8 (which fixed another issue), the doc title changed to where the git commit was quoted.

I'll assume it wasn't intentional and will provide a PR to skip inclusion of the single quotes. Please feel free to reject the PR if the behavior was intentional or the results desired.

PR #744, opened 2018-11-13
Escape single quotes in git log command

deoren:
Escape the single quotes in the git log format command so that they are not interpreted literally and included with the doc title when output files are generated.

I'm assuming that the behavior is not intentional, but please feel free to reject this PR if it is.

closes rsyslog/rsyslog-doc#743

### DIFF
--- a/source/conf_helpers.py
+++ b/source/conf_helpers.py
@@ -78,7 +78,7 @@ def get_next_stable_version():
 def get_current_commit_hash():
     """Return commit hash string"""
 
-    commit_hash = subprocess.check_output(['git', 'log', "--pretty=format:'%h'", 'HEAD', '-n1']).decode("utf-8")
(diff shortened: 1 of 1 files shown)


deoren:
I was trying to leave the double-quotes in, just escape them, but that doesn't appear to be sufficient to fix this. Replacing with escaped double-quotes instead. Will update branch in a moment.

rgerhards:
> I'm assuming that the behavior is not intentional,

I would also assume so.

> Will update branch in a moment.

pls let me know when done

deoren:
Sorry for the delay. I ended up building from a cache that had a prior edit and got myself mixed up. Realized what happened and performed a full rebuild to confirm the results via:

- `sphinx-build -E -b html source build`
- `sphinx-build -E -b epub source build`

rgerhards:
@deoren just as a reminder, the CI system builds a version for each PR: https://build.rsyslog.com/#/builders/111/builds/134  (last step, `preview url`)

PR #748, opened 2018-11-15
Add the TCP_Framing options ("traditional" or "octet-counted")

hynd:
Couldn't find anywhere in the docs what the two option words were.


rgerhards:
thx for the PR!

PR #755, opened 2018-12-10
Add hostnames examples in TLS.PermittedPeer and a note in TLS.AuthMode

trourance:
Based on the comments found at the bottom on this page rsyslog/rsyslog#435 and the problems encountered in this discussion http://rsyslog-users.1305293.n2.nabble.com/rsyslog-RELP-and-TLS-creating-the-certificates-tp7594374.html, I've modified the TLS section of omrelp module.


rgerhards:
looks like the reported issue was already solved in current master; could originally  merge due to CI failure, so can now close.

PR #766, opened 2019-01-03
omclickhouse: add documentation for new parameters

PascalWithopf:
Documentation for following parameters were added
1. allowUnsignedCerts, which allows connections with
servers, without checking if the certificates are authentic.
2. healthCheckTimeout, which sets the timeout for the
availability check for the clickhouse server.


davidelang:
is there an adiscon hosted repo that has all the versions in it? if not, it 
would probably be a good idea to have one.

Most people should use the PPA, but occasionally there is a reason to try older 
versions (troubleshooting if nothing else)

David Lang

PR #787, opened 2019-02-25
imfile: added doc for stateFileAlwaysRemove

alorbach:


rgerhards:
PR has been replaced, so this option will not exist.

PR #788, opened 2019-02-25
add note to 'resubmitOnFailure' option.

n2yen:
- messages rejected due to exceeding maximum configured message size will be
   automatically dropped.


n2yen:
In reference to [3515](https://github.com/rsyslog/rsyslog/pull/3515)

PR #810, opened 2019-04-25
syntax error in sysklogd_format

eagleoflqj:
Before I made changes, all the 4 lines rendered wrong in both https://www.rsyslog.com/doc/v8-stable/configuration/sysklogd_format.html and the html I generated with sphnix.
Although line 27 and 104 rendered correctly in github preview, I think the html in official website should be ultimate version.


rgerhards:
thx!

PR #827, opened 2019-06-20
Fix typo

atc0005:
Are minor tweaks like this welcome?


rgerhards:
Definitely! Thx for it.

PR #840, opened 2019-08-08
Fix gelf documentation

elcamlost:
Now gelf forwarding tutorial claims that you cannot use omwd tcp and graylog together due nullbyte as frame delimiter.
It is false. Frame Delimiter is configurable in rsyslog since 2017. https://github.com/rsyslog/rsyslog/issues/1602

Also, timestamp field in modern gelf should be number, not string. So quotes around timestamp value is against gelf standard.

This PR is fixing both issues.


rgerhards:
thx for the PR, much appreciated!

PR #851, opened 2019-10-15
fix display of codeblock

igalic:
use a code-block, rather than backticks, and actually indent where needed, so this example becomes clear.


rgerhards:
thx for the PR!

PR #862, opened 2019-11-22
Add documentation for omhttp submittal to loki

dmolik:


dmolik:
in support of rsyslog/rsyslog#3976 and rsyslog/rsyslog#3982

PR #868, opened 2019-11-27
Realigning doc to source code

paride5745:
In `rsyslog/runtime/glbl.c`, the parameter is defined as `reverselookup.cache.ttl.default` instead of `reverselookup.cache.default.ttl`


rgerhards:
good find - many thx!

PR #876, opened 2019-12-20
omhiredis: add 'set' mode, 'expiration' parameter, update examples

frikilax:
- added set mode
- added expiration parameter
- added example(s) for set, with and without expiration
- updated examples for clarity


frikilax:
Missing 'dynakey' parameter, adding it

frikilax:
All good

PR #887, opened 2020-02-21
ossl: Add doc hint for chained certificates

alorbach:
see also: https://github.com/rsyslog/rsyslog/pull/4182
see also: https://github.com/rsyslog/rsyslog/issues/4170


rgerhards:
I think there is no need to pint this out specifically for the openssl driver - GnuTLS always supported it.

PR #893, opened 2020-04-13
Correction for escapeLF.replacement

trantor:
Wrong parameter referenced in imfile.rst for escapeLF.replacement


rgerhards:
thx for the PR ... and sorry it took so long to merge.

PR #907, opened 2020-09-13
Add documentation for imhttp input module

n2yen:
documentation for imhttp input module PR: https://github.com/rsyslog/rsyslog/pull/4301

### DIFF
--- a/source/configuration/modules/imhttp.rst
+++ b/source/configuration/modules/imhttp.rst
@@ -1,0 +1,316 @@
+*************************
+imhttp: http input module
+*************************
+
+===========================  ===========
+**Module Name:**             **imhttp**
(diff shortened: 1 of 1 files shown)


rgerhards:
@n2yen It would be great if you could also add a ruleset processing the messages to the samples. The point is if they can be processed "just regularly" -- I would assume they usually do not have a valid syslog format. I may be wrong.

n2yen:
hi @rgerhards , yes, i will add some more ruleset processing samples there.
In regards to if the messages need to be in valid syslog format - doesn't have to be.
There are essentially 3 modes:
- treat each LF as a separate message.
- ignore LF and submit entire post data as a single message.
- syslog style framing.

I'll try and clarify this in the samples.

PR #910, opened 2020-10-06
omfwd doc for new feature for PR4440

Dinesh-Ramakrishnan:
Adding documentation for the new rate limit option
of omfwd for rate limiting syslog messages sent to the
remote server

ratelimit.interval:
Specifies the rate-limiting interval in seconds.
Default value is 0, which turns off rate limiting.

ratelimit.burst
Specifies the rate-limiting burst in number of messages.

Reference:
https://github.com/rsyslog/rsyslog/issues/4423
https://github.com/rsyslog/rsyslog/pull/4440

Signed-off-by: Dinesh-Ramakrishnan <pr.dinesh@gmail.com>

### DIFF
--- a/source/configuration/modules/omfwd.rst
+++ b/source/configuration/modules/omfwd.rst
@@ -315,7 +315,7 @@ RebindInterval
 
 Permits to specify an interval at which the current connection is
 broken and re-established. This setting is primarily an aid to load
-balancers. After the configured number of batches (equals roughly to 
+balancers. After the configured number of batches (equals roughly to
 messages for UDP traffic, dependent on batch size for TCP) has been
 transmitted, the current connection is terminated and a new one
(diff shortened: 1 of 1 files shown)


rgerhards:
I had only a quick glimpse at the PR by now. From that I have the impression that over-the-ratelimit messages will be discarded. If so, that should definitely be documented. For senders, IMHO, many people expect that message queueing happens and so no message is lost but a steady max state is reached. I agree that it is useful to be able to discard, but the distinction and consequences should be clear.

rgerhards:
@Dinesh-Ramakrishnan while discarding the message is in-line of what we do at the input side, I think it would be good to add an explicit note. Are you up for this?

Dinesh-Ramakrishnan:
Sorry, I missed to response.  Yes, we can mention that in the doc.

PR #916, opened 2021-02-15
Documentation for ConErrSkip option.

bukatlib:
Add documentation for ConErrSkip option.


bukatlib:
See also: https://github.com/rsyslog/rsyslog/pull/4532

PR #918, opened 2021-02-23
Added gztool as a tool for extracting data when veryRobustZip is set to off

circulosmeos:


rgerhards:
after a bit of thinking, this sounds like a fair merge ;-)

PR #924, opened 2021-04-09
rainerscript/global.rst: control characters escaping options

Ubusubus:
Copied global control character escaping options from legacy format to rainerscript/global doc and changed to advanced format.


rgerhards:
thx for the PR!

PR #927, opened 2021-04-21
Updated "High Database Rate" -> converted example configuration from legacy format

stevleibelt:
I stumbled upon this section two days ago.
I tired to configure my rsyslog that it can handle a table lock while my [housekeeping](https://github.com/bazzline/rsyslog_mysql_housekeeping) script is removing old entries from the `Syslog`.`SystemEvents` database table.

With the existing configuration, I was not able to make it work. With my pushed and updated configuration, all is working as expected.
Rsyslog is telling me that the database server is not reachable, there are now files with the name in my `$WorkingDirectory` and if my database server is back online, the files are removed and I see the missing entries in my table.

### DIFF
--- a/source/configuration/index.rst
+++ b/source/configuration/index.rst
@@ -34,6 +34,7 @@ Note that **configurations can be built interactively** via the online
    cryprov_gcry
    dyn_stats
    lookup_tables
+   percentile_stats
 
 `Configuration file examples can be found in the rsyslog
 wiki <http://wiki.rsyslog.com/index.php/Configuration_Samples>`_. Also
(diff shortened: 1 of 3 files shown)


stevleibelt:
I don't get the errors from the build. I've "just" copied the original changelog line and replaced the name with my name 🤔 .

rgerhards:
Sorry, I overlooked this PR. The issue was that you copied a reference instead of plain text. I have fixed the nit (just use the plain name). 

Oops... and something seems to have wrong while doing so... let me check. ;-)

PR #934, opened 2021-06-24
omhttp: update inverted examples for kafka and loki

johanfleury:
The documentation for `lokirest` introduced in #862 inverted examples between kafkarest and the newly added lokirest.


rgerhards:
could you rebase to current master? For some reason, CI seems not to correctly apply the merge commit including current master. I updated sphinx, and sphinx saw an error it previously ignored, I fixed it, fix is in master, but re-running here does not help (what it should...).

rgerhards:
looks like it healed...

rgerhards:
oh, and BTW: thanks for the PR! :-)

PR #940, opened 2021-08-15
Doc on percentile_stats module, and percentile_observe function

n2yen:
reference: https://github.com/rsyslog/rsyslog/pull/4612

### DIFF
--- a/source/configuration/index.rst
+++ b/source/configuration/index.rst
@@ -34,6 +34,7 @@ Note that **configurations can be built interactively** via the online
    cryprov_gcry
    dyn_stats
    lookup_tables
+   percentile_stats
 
 `Configuration file examples can be found in the rsyslog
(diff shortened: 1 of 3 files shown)


davidelang:
it says that the fifo buffer is rounded to the nearest power or 2, I think it should be rounded UP to the nearest power of 2, you don't want to end up with a buffer smaller than what the person specified. (the power of 2 should be an optimization, not something that the admin needs to think about)

n2yen:
> it says that the fifo buffer is rounded to the nearest power or 2, I think it should be rounded UP to the nearest power of 2, you don't want to end up with a buffer smaller than what the person specified. (the power of 2 should be an optimization, not something that the admin needs to think about)

thanks @davidelang for pointing it out. It is implemented as you described. I'll update the wording!

PR #941, opened 2021-08-16
fixing up merge issues (maybe)

rgerhards:
see #927 for original PR


rgerhards:
@stevleibelt please note that there were some syntax (no main_queue param in action object) and semantic (first sample did not use a global main queue) problems. I have fixed these as well.

PR #950, opened 2021-11-01
Fix link: monitorware.com -> adiscon.com

tobiaspal:
The http://www.monitorware.com/en/syslog-enabled-products/ URL does not work, I found https://www.adiscon.com/syslog-enabled-products/ which seems to be the replacement.


rgerhards:
Thx for the PR, much appreciated.

PR #971, opened 2022-07-12
omkafka: added dynaKey available version

supery-chen:
The **DynaKey** param is available since [v8.1903](https://github.com/rsyslog/rsyslog/pull/3350), but it's not mentioned in the [document](https://www.rsyslog.com/doc/master/configuration/modules/omkafka.html#dynakey)


rgerhards:
thx for the PR

PR #1004, opened 2023-02-22
fix: backslash escape

ipetko96:
again, too many `\` caused double backslash in the documentation


rgerhards:
and double backslash is absolute the right thing to get a single backslash inside a constant string ;-)

rgerhards:
No problem at all :-) -- but the other PR was right! Thx for checking and correcting.

PR #1007, opened 2023-05-22
Fix double quotes, add omprog send mail example

paulfertser:
omprog: add example script to run an app for each message
    
    This adds a simple POSIX shell script that can be used to execute an
    program for every log message (as the legacy ^ action).
    
    Since the example shows how to use it to send mail, a reference is added
    to the ommail page too.


paulfertser:
The build failed probably because it tried to fetch too early? I do not see how I can retrigger it.

PR #1017, opened 2023-10-03
Document libcapng.enable global option

Cropi:
Document https://github.com/rsyslog/rsyslog/pull/5234


rgerhards:
Pls ignore the CI fail - it looks like we need to do some work on the CI definitions :-)

rgerhards:
@Cropi oops, I was wrong. there is indeed an issue that sphinx does not like. A fix would be appreciated.

Cropi:
I fixed it. Actually the problem was in the imhredis part. Thanks for pointing it out.

rgerhards:
@Cropi The CI fail is a bit strange, especially as imhiredis never showed issues. But in the new param the underline is too long. Maybe sphinx is confused. I'll see if I can do a fixup commit and, if so, hope that resolves the situation.

Issue #394, opened 2017-11-24
documentation is hard to use and badly structured


grinapo:
I sincerely apologise for the seemingly negative attitude of mine beforehand.

I have been suffering the rsyslog documentation for years now. I tried to be calm and positive, hoping that it'll get better as the program gets wider use, but everytime when I have to go back to rsylog I realise how painful it is. 

My main problem is structure and clarity. The whole documentation makes a complex dance around various formats the config file may use, scattered everywhere all kind of syntax, often mixed. The config base structure is never mentioned, and the order of the elements in the TOC is seemingly random.

Let's change to constructive (without actually rewriting the docs since I only use rsyslog when I am forced to, mostly due to this problem). What do I have in mind?

Like first start with **describing the alternative formats**, clearly defining whether it is obsolete, or alternative, or preferred. This part should actually take real life in account, like I've never seen a distro using "rainserscript" despite that the doc suggest all else is obsolete. Except other parts say it's the opposite, as rainerscript is slow and undesirable. Or else.

Then describe the formats **separately**. Make all documentation self-contained and whole, so if I am not interested in, say, rainerscript, I do not have to browse through its documentatio nfor hints related to old syslog format.

Doing that you may start with structure: what is the structure of the config format? and I mean full structure.  Like there is no description for old-syslog format like:

"The general format is:
[...] filter
[...] filter
[...]
[...] action
& [...] action
& [...] action
[...]

where filter is:
....

and within the elements are
.....

where action is:
..."

So right now it's almost impossible to see whether it's posisble to use more than one filters, and they're semicolon separated, or new line separated, or both, or neither; whether they are ANDed or ORed together; whether the filename element is compulsory or not, or it may be something else than filename (tilde, forgeign host etc); whether actions may be put on second line, or lines, and how and why.

And do not try to cram it into one page and one paraghraph. Show the structure, and link to separate pages describing the elements. 

And this is only just one example how to describe the **whole** syntax **clearly**, and after that there could be the list of various examples.

The current documentation feels like a tutorial, which shows usage examples (and badly that) without specification and syntax. Intermixing the formats. And mixing all kinds of usage. Not considering real-life.

I can give you ideas, structure. I won't write documentation, sorry. I can check it and can offer insights, if I must. And you may close this issue if you feel you can't help it and there are no other volunteers to get it done.

rgerhards:
[...]

PR #1024, opened 2023-11-10
clarify that "advanced" format is the way to go

rgerhards:
We no longer recommend "basic" format for simple things for beginners.

see also https://github.com/rsyslog/rsyslog-doc/issues/394


rgerhards:
did not find community approval

PR #1035, opened 2024-01-18
Update gelf_forwarding.rst

kboenke:
Puts together required pieces of configuration for traditional priority-based filters.


rgerhards:
We cannot merge this, as the required framing can not be set by traditional format.

PR #1043, opened 2024-02-21
Document supported values for securepath parameter in template property statements

yrro:


rgerhards:
my pleasure and thx to you!

PR #1045, opened 2024-02-26
Note how to avoid path traversal attacks with omfile

yrro:
omlibdbi warns users about SQL injection attacks; it seems wise to do the same for path traversal attacks with omfile.


rgerhards:
thx for the PR!

PR #1048, opened 2024-05-16
Added two document files; KEYWORDS.csv and METADATA.csv

egberts:
Added two document files; KEYWORDS.csv and METADATA.csv

Was working on railroad diagrams and these files came in handy.

### DIFF
--- a/contrib/KEYWORDS..csv
+++ b/contrib/KEYWORDS..csv
@@ -1,0 +1,785 @@
+
+name, function type, module, variable type, default value, mandatory, alternate, introduced in, obsoleted by, description
+`action.copyMsg`, action, builtin
+`action.errorfile`, action, builtin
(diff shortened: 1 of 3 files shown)


rgerhards:
I have some concerns adding these files because there is no process that keeps them current. I propose to better add a script that generates them based on the version checked out from git.

egberts:
If you can figure out a way to extract the environment name string that is computed dynamically, then yes we can. But it would require some kind of an Abstract Syntax Tree reconstructor to just to extract the environment name.  

Just mark it as developer internal doc and be done with.

rgerhards:
non maintainable, closing this

egberts:
Might explains why documentation is hard to find, notably with environment variables.

Anyway, for the general peruser, you can find the environment variables (and other things) here:

https://wiki.gentoo.org/wiki/Rsyslog

PR #1051, opened 2024-06-08
Fix typos

bobidle:
Again, notice some words that don't look correct. Beside that i changed some British English words to the American English form, like `specialised` to `specialized`, as American English forms seems a little bit more common in the rsyslog docs.


rgerhards:
thx again for the pr!

PR #1067, opened 2024-10-24
Fix typos

bobidle:
I possible found some more typos and tried to fix them :crossed_fingers:


rgerhards:
Thx for the patch, much appreciated!

PR #1076, opened 2025-01-27
document new omfwd paramter name aliases

rgerhards:
see also: https://github.com/rsyslog/rsyslog/pull/5540/files


rgerhards:
@davidelang FYI

